PR #47, opened 2024-02-20
Prepare icon kit release

Haberkamp:
## What?

I updated the changelog to use the new changeset workflow.

## Why?

This way we can release the version update and do it through our new automated approach.

## How?

Removed the old changelog and replaced it with a changeset.

### DIFF
--- a/packages/icon-kit/package.json
+++ b/packages/icon-kit/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopware-ag/meteor-icon-kit",
-  "version": "5.2.1",
+  "version": "5.2.0",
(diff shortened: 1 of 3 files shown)


packages/icon-kit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopware-ag/meteor-icon-kit",
-  "version": "5.2.1",
+  "version": "5.2.0",
```
jleifeld, line 3: Is this wanted?

jleifeld (changes requested):

jleifeld (approved):

jleifeld:
Merge after https://github.com/shopware/meteor/pull/46

PR #51, opened 2024-02-21
feat: add developer-portal workflows

bojanrajh:
## What?
3 new workflows for communicating with the DevHub:
 - [X] `developer-portal-healthcheck` - triggers a new [preview build](https://github.com/shopware/developer-portal/blob/main/.github/workflows/external-healthcheck.yml) in `developer-portal` for every PR targeting `main` branch in this repo
 - [X] `update-healthcheck` - needed to update the PR status check from the previous workflow - will be replaced with a dedicated GitHub app to avoid usage of the PAT
 - [X] `deploy-developer-portal` - triggers a new [production build](https://github.com/shopware/developer-portal/blob/main/.github/workflows/checkout-test-build-deploy.yml) in `developer-portal` whenever the `main` branch has changed
 - [ ] `DEV_HUB_PERSONAL_ACCESS_TOKEN` - PAT, tmp

## Why?
So we can automate DevHub actions.

## How?
By adding a few workflows.

## Testing?
Already used by `shopware/docs`, `shopware/frontends`, `shopware/release-notes`.

## Anything Else?

### DIFF
--- a/.github/workflows/developer-portal-healthcheck.yml
+++ b/.github/workflows/developer-portal-healthcheck.yml
@@ -1,0 +1,19 @@
+# This workflow triggers a Healthcheck workflow in developer-portal for every PR targeting main branch.
+name: Healthcheck
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  create-healthcheck:
+    uses: shopware/developer-portal/.github/workflows/healthcheck.yml@main
+    with:
+      owner: ${{ github.repository_owner }}
+      repo: ${{ github.event.repository.name }}
+      branch: ${{ github.event.pull_request.head.ref }}
+      sha: ${{ github.event.pull_request.head.sha }}
+    secrets:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      PERSONAL_TOKEN: ${{ secrets.DEV_HUB_PERSONAL_ACCESS_TOKEN }}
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules/
 .turbo
 .env
 chrome/
+.idea
```
Haberkamp, line 5: I know a bit nit picky. Could you do that in a separate commit?

Haberkamp (changes requested):
One workflow seems to be broken.

bojanrajh:
> One workflow seems to be broken.

It will work once the `DEV_HUB_PERSONAL_ACCESS_TOKEN` secret is set.

.github/workflows/developer-portal-healthcheck.yml
```diff
@@ -0,0 +1,19 @@
+# This workflow triggers a Healthcheck workflow in developer-portal for every PR targeting main branch.
+name: Healthcheck
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  create-healthcheck:
```
Haberkamp, line 10: I would rename this workflow and job to something that tells me this job creates a preview build. I understand a healthcheck as an Rest endpoint that you ping and you get a response back that the system is running without any issues.

.github/workflows/update-healthcheck.yml
```diff
@@ -0,0 +1,38 @@
+# This workflow is triggered from developer-portal and updates the healthcheck status in the PR.
+name: Update healthcheck
```
Haberkamp, line 2: Here the same as above. I would rename this to something like "Update developer portal preview"

Haberkamp (changes requested):

bojanrajh:
This PR is not needed with new `DevHub connector` GitHub app, therefore I'm closing it.

PR #56, opened 2024-02-22
Sync component library

Haberkamp:
## What?

Syncing latest changes to the monorepo

## Why?

We need to sync this a last time.

## How?

Merging changes into this repository

## Testing?

Please check the diff and maybe do some manual testing.


packages/component-library/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@shopware-ag/meteor-component-library",
```
Haberkamp, line 2: @jleifeld could you check what's correct here?

packages/component-library/tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+<<<<<<< HEAD
```
Haberkamp, line 2: @jleifeld could you try to resolve this merge conflict? I don't know that the right config should be.

packages/component-library/package.json
```diff
@@ -13,6 +14,23 @@
     "start:storybook": "serve ./storybook-static -p 6006 --no-request-logging",
     "test:storybook": "test-storybook --url http://127.0.0.1:6006",
     "docker": "docker run -v $PWD/../../:/tests:delegated -w /tests -it --rm --ipc=host mcr.microsoft.com/playwright:v1.20.0-focal /bin/bash"
+=======
+  "version": "3.0.0-alpha1",
```
Haberkamp, line 18: We have to pay attention on how we proceed here. The [changesets documentation](https://github.com/changesets/changesets/blob/main/docs/prereleases.md) states that one must pay attention when doing pre-releases.

packages/component-library/package.json
jleifeld: What do you mean exactly?

packages/component-library/package.json
jleifeld: I think for the initial merge it should work because it is already published. Additional prerelases should be released in a separate branch

jleifeld:
Playwright Integration Tests were skipped

Haberkamp:
The playwright tests only run once the other tests succeed, otherwise they're skipped. This way we save some computing resources.

PR #57, opened 2024-02-23
Re-embed Icon kit @ DevHub

bojanrajh:
## What?
 - [X] provides `docs:env`, `docs:link` and `docs:preview` shortcuts/commands for Icon kit
 - [X] migrates paths to the new directory structure
 - [X] adds mete generation script for fetching tags from Figma
 - [ ] fixes meta generation script (?)

## Why?
So we can switch Icon kit docs to the new monorepo. 

## Testing?
There's a healthcheck workflow in https://github.com/shopware/meteor/pull/51 .

## Anything Else?
We can remove the workflow from `developer-portal` if updating `meta.json` becomes part of the release process (I see it's [part of the repo now](https://github.com/shopware/meteor/commits/main/packages/icon-kit/icons/meta.json)).


.github/scripts/docs-after-clone.sh
```diff
@@ -0,0 +1,18 @@
...
+[[ -z "$1" ]] && echo "Missing working directory argument" && exit 1
+
+#echo "Installing root npm"
+#pnpm -C $1 i
+
+#echo "Running root build"
+#pnpm -C $1 --filter "*/meteor-icon-kit" build
+
+#echo "Building figma"
+#pnpm -C $1 --filter "*/meteor-icon-kit" start
```
Haberkamp, line 13: Can those be removed?

Haberkamp (changes requested):

.github/scripts/docs-after-clone.sh
bojanrajh: I'll delete the whole file because the build step is not needed anymore and we can control some stuff from the `docs.yml`.

fabianhueske:
Hi @bojanrajh 👋 Is this PR still needed?

PR #62, opened 2024-02-26
NEXT-33669 - implement cms blocks for sdk

jleifeld:
## What?

Add the possibility to add CMS blocks via the Admin SDK

## Why?

Currently it is only possible with the Manifest file which is much more complicated

## How?

Add the types


jleifeld:
Do not merge yet!

Missing things:
- [ ]  Docs
- [ ]  Storefront implementation checks

seggewiss (approved):

PR #64, opened 2024-02-26
Fixup sdk

seggewiss:


jleifeld (approved):

packages/admin-sdk/src/_internals/utils.ts
```diff
@@ -1,4 +1,8 @@
+/**
+ *
+ */
```
Haberkamp, line 3: Is there a particular reason for this to exist? Has this any benefit?

Haberkamp (changes requested):

packages/admin-sdk/src/_internals/utils.ts
seggewiss: Good catch! Thank you 😊

Haberkamp (approved):

PR #71, opened 2024-02-27
NEXT-33861 - improve selector

jleifeld:
## What?

Update the selectors in datasets to add more functionalities

## Why?

The selectors in the datasets was using Lodash "get" and was very limited

## How?

Replace it with better alternative

## Testing?

See tests


jleifeld (commented):
Looks good. Only thing missing is a changelog :)

Haberkamp (approved):

PR #86, opened 2024-03-07
Move example app into example directory

Haberkamp:
## What?

I turned turned of the ability to publish the example app to the npm registry. I also moved the package into an examples folder.

## Why?

The package is not intended to be published to npm.

1. By having two directories it makes it immediately clear what is a package I can install via npm and what is a code example. 
2. Using a separate folder makes it scale better. Using the old structure, when we have multiple code examples it makes it harder to navigate to the actual packages you want. Most of the time you don't care about the examples but only on the packages. Hiding them makes it easier to navigate to those packages.

The example app is still able to consume the other packages stored in the monorepo.

## How?

To prevent the package from publishing I set the `private` property to `true`.

## Testing?

You can checkout this repository, go into the example directory and manually test that the package still works.

## Anything Else?

When this PR gets merged I'll also delete the package from npm.


jleifeld:
When I execute `pnpm run --filter meteor-admin-sdk-example-app dev` it is not working:

jleifeld (approved):

seggewiss:
We can define our own standard, with an example folder. Other projects use an apps folder for such things. If we stay with example the folder in it does not need to contain example again in the name.

seggewiss (commented):
We can define our own standard, with an example folder. Other projects use an apps folder for such things. If we stay with example the folder in it does not need to contain example again in the name.

Haberkamp:
Yes, other projects use the `apps/` folder for things they deploy to the web. Like their homepage or their docs. 

But they make a distinction between apps and code examples. They put their code examples into an `examples/` folder. They don't deploy their code examples, they only offer them to devs to learn that tool or to get started quicker. 

The projects below follow that structure:
* https://github.com/nuxt/nuxt
* https://github.com/vercel/next.js
* https://github.com/shopware/frontends
* https://github.com/vercel/turbo

If we put code examples into an `apps/` folder we have the same problem. We label two different things as the same thing.

jleifeld (approved):

PR #92, opened 2024-03-13
NEXT-25907 - implement character count

jleifeld:
## What?

sw-textfield and sw-textarea should have a character count as a feature

## Why?

If a error gets thrown that the text content is too long a count indicator can be helpful for the user to write a text with the correct length

## How?

Implement it as a optional prop

## Testing?

Tested with Storybook interaction tests

## Screenshots (optional)
interaction-tests-form-sw-text-field--visual-test-character-count-with-hint-snap


packages/component-library/__snapshots__/interaction-tests-navigation-sw-tabs--visual-test-render-context-tab-with-error-snap.png
Haberkamp, line 1: Text "More" text should be red. I missed that while updating the component. I'll noted that down and fix that once I'll get to work on the tokens again.

packages/component-library/src/components/form/sw-button/sw-button.vue
```diff
@@ -7,6 +7,7 @@
     class="sw-button"
     :class="buttonClasses"
     v-bind="$attrs"
+    role="button"
```
Haberkamp, line 10: This is an accessibility violation. It may look like a button but it works like a link, because it is one.

packages/component-library/src/components/form/sw-text-field/sw-text-field.stories.ts
```diff
@@ -26,7 +26,7 @@ export default {
     <div>
       <sw-text-field
         v-bind="args"
-        :modelValue="currentValue"
+        v-model="currentValue"
         @inheritance-restore="inheritanceRestoreWrapper"
         @inheritance-remove="inheritanceRemoveWrapper"
         @change="onChange">
```
Haberkamp, line 32: If the `v-model` is used here than we can get rid of the 'change` event listener line 32.

Haberkamp (changes requested):
It looks like the limit for the `sw-text-field` is not working.

https://meteor-component-library-git-next-259-e2bddd-shopware-frontends.vercel.app/?path=/story/components-form-sw-text-field--default-story&args=maxLength:12;modelValue:This+text+is+over+the+limit

packages/component-library/src/components/form/sw-text-field/sw-text-field.stories.ts
jleifeld: The change event listener is needed for showing the event inside the Storybook events

packages/component-library/src/components/form/sw-button/sw-button.vue
jleifeld: You're correct. I added it because some tests were failing because they are looking for a button. But I will adjust the tests then

jleifeld:
> It looks like the limit for the `sw-text-field` is not working.
> 
> https://meteor-component-library-git-next-259-e2bddd-shopware-frontends.vercel.app/?path=/story/components-form-sw-text-field--default-story&args=maxLength:12;modelValue:This+text+is+over+the+limit

I checked this behavior in different other applications (e.g. Slido, etc.) and component libraries. There it is possible to go over the limit so that it is only a visual helper. Other downside could be that after implementing it, it could remove content which would existed beforehand.

@Weltraumakustik what do you think should be the correct behavior?

jleifeld:
> > It looks like the limit for the `sw-text-field` is not working.
> > https://meteor-component-library-git-next-259-e2bddd-shopware-frontends.vercel.app/?path=/story/components-form-sw-text-field--default-story&args=maxLength:12;modelValue:This+text+is+over+the+limit
> 
> I checked this behavior in different other applications (e.g. Slido, etc.) and component libraries. There it is possible to go over the limit so that it is only a visual helper. Other downside could be that after implementing it, it could remove content which would existed beforehand.
> 
> @Weltraumakustik what do you think should be the correct behavior?

Limit was added with behavior most natively to the maxLength behavior of a default HTML Input

PR #96, opened 2024-03-15
Add nuxt example app

Haberkamp:
## What?

I added a Nuxt SSR example project.

## Why?

This helps us to see wether or not our examples are SSR compatible.

## How?

I added a new project.

## Testing?

There is an integration test to see if the application is working correctly.

## Anything Else?

There are several more things I did:
1. I added the `examples/` directory to the readme of the project
2. I included the tokens in the global css file for the meteor components
3. I fixed the import path in the readme of the token project


examples/nuxt-app/playwright.config.ts
```diff
@@ -0,0 +1,72 @@
...
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] },
+    },
+
+    {
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
+    },
+
+    /* Test against mobile viewports. */
```
seggewiss, line 45: Can these comments go?

jleifeld (changes requested):
Looks good. I would also add a screenshot test (natively supported in Playwright) to make sure that the component rendering works correctly: https://playwright.dev/docs/test-snapshots

Then we see if we mess up with CSS and other stuff

examples/nuxt-app/playwright.config.ts
Haberkamp: Sure!

seggewiss (approved):

jleifeld (approved):

PR #101, opened 2024-03-22
NEXT-34522 - rename meteor components to mt

jleifeld:
## What?

Everything `sw-` and `SwSomething` have to be renamed to `mt-` and `MtSomething`

## Why?

Conflicting names with the Shopware platform

## How?

Rename it

## Testing?

Test should still work


Haberkamp (approved):

Haberkamp (changes requested):
This is a breaking change. So it must also be reflected correctly in the changeset.

jleifeld:
> This is a breaking change. So it must also be reflected correctly in the changeset.

Our external API is not affected by the changes. I added the old export names to avoid breaking changes: https://github.com/shopware/meteor/pull/101/files#diff-ce485795d54caece4b1ef8a6ea674174ceedc56812466358c10d0254d412ae99R53

Changing internal CSS and Component Names shouldn't lead to breaking changes. Because importing the objects normally works the same as before

Haberkamp (approved):

PR #106, opened 2024-03-27
NEXT-34673 - upgrade to storybook 8

Haberkamp:
## What?

I upgraded storybook to the latest version.

## Why?

The upgrade brings several improvements with it. Interesting to us is the better Vue.js support as well as faster builds to shorten the execution time of the pipeline.

## How?

I ran the automigrations provided by storybook and checked wether or not the majority of stories still works.

## Testing?

Please do some manual testing through the deployment build.


jleifeld (changes requested):
Everything seems still working. But the improvements, like auto-doc generation isn't active yet. See:

Bildschirmfoto 2024-03-27 um 13 46 01

Normally this should be possible with Storybook 8 (https://storybook.js.org/blog/first-class-vue-support-storybook-8/). 

So it would be cool if we can opt-in this feature, see the docs:
https://storybook.js.org/docs/get-started/vue3-vite?ref=storybookblog.ghost.io#using-vue-component-meta

Because then every developer have much better experience using the components 🙂

jleifeld (approved):

PR #111, opened 2024-04-03
NEXT-34341 - add filter

Haberkamp:
## What?

Adding filters to the data table.

## Why?

Our users work with large sets of data. The filters make it easier for them to work with that data.

## How?

I added filters to the data table.

## Testing?

You can test it out via the preview deployments below. There is also a sufficient number of tests for the new filters.


packages/component-library/__snapshots__/interaction-tests-form-mt-colorpicker--visual-test-change-colorpicker-color-snap.png
jleifeld, line 1: Why did the font and background color changed?

packages/component-library/__snapshots__/interaction-tests-form-mt-switch--visual-test-bordered-error-snap.png
jleifeld, line 1: Just for me: why did this changed?

packages/component-library/__snapshots__/interaction-tests-overlay-mt-popover--visual-test-render-child-view-snap.png
jleifeld, line 1: Padding below search looks to small in the new one

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.stories.ts
jleifeld, line 1: I would add the filters to the default story so that it is visible for anyone

jleifeld, line 1: In general: is the design currently finished? If not, I would suggest to add information about this for external developers

packages/component-library/src/components/form/_internal/mt-base-field/mt-base-field.vue
```diff
@@ -38,7 +38,10 @@
 
     <slot name="error" />
 
-    <div class="mt-field__hint-wrapper">
+    <div
+      v-if="!!$slots['field-hint'] || !!$slots['field-hint-right']"
```
jleifeld, line 42: The v-if with slots is critical. We shouldn't use this anymore because it won't get updated when the slot gets filled later. In the case of a form field this could be a valid case when a hint gets added later on.

I would prefer :empty in CSS

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
```diff
@@ -1446,6 +1592,107 @@ export default defineComponent({
       }
     };
 
+    function removeFilter(id: string) {
+      const updatedFilters = props.appliedFilters.filter((filter) => filter.id !== id);
+
+      emit("update:appliedFilters", updatedFilters);
```
jleifeld, line 1598: Personal thing: I would prefer events which are more explicit, e.g. in this case just a event which tells that a filter was removed and the id.

The reason for this is that with sending all filers directly it is harder from the outside to get the changes. Because in most cases you won't to replace the filter data in the parent component with the newly emitted one because it is very likely that some reactivity deps wouldn't get triggered that way. So the parent component need to create a diff between the new filters from the event and update them on the real data then.

So it would be easier to maintain and less error prone if they would be more explicit. The dumber the base component, the easier to maintain.

jleifeld (commented):
Big feature 👍  Looks good in general. Have only some questions and smaller feedbacks

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.stories.ts
Haberkamp: What do you want to achieve / convey with that information or differently formulated what do want to put in that information?

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.stories.ts
```diff
@@ -334,6 +337,10 @@ export default {
         showStripesValue: boolean;
         enableOutlineFramingValue: boolean;
         enableRowNumberingValue: boolean;
+        // TODO: add better types for filters
```
jleifeld, line 340: Missing todo

jleifeld (approved):

PR #114, opened 2024-04-03
NEXT-34340 - Implement toasts

seggewiss:
## What?
Implement toast component into component library.


packages/component-library/src/components/feedback-indicator/mt-toast/mt-toast-notification.vue
```diff
@@ -0,0 +1,376 @@
...
+    required: false,
+    default: NaN,
+  },
+});
+
+const classes = computed(() => {
+  return {
+    "mt-toast-notification__positive": toast.value.type === "positive",
+    "mt-toast-notification__critical": toast.value.type === "critical",
+    "mt-toast-notification__informal": toast.value.type === "informal",
```
Haberkamp, line 88: According to the BEM methtod variants are used with `--`. When we want to target a child we use the underscores.

packages/component-library/src/components/feedback-indicator/mt-toast/mt-toast.vue
```diff
@@ -0,0 +1,245 @@
...
+const rightToasts = computed(() => {
+  return toasts.value.filter((t) => {
+    return t.action || t.type === "critical" || t.dismissible;
+  });
+});
+
+const classes = computed(() => {
+  return {
+    "mt-toast__collapsed": !isHovered.value,
+    "mt-toast__expanded": isHovered.value,
```
Haberkamp, line 85: Same thing as in the other comment.

packages/component-library/src/components/feedback-indicator/mt-toast/mt-toast.vue
```diff
@@ -0,0 +1,245 @@
...
+}
+
+function onMouseLeave() {
+  hoverTimeoutId.value = window.setTimeout(() => {
+    isHovered.value = false;
+  }, 500);
+}
+</script>
+
+<style lang="scss">
```
Haberkamp, line 112: Can we used scoped styles here?

packages/component-library/src/index.ts
```diff
@@ -94,4 +96,6 @@ export {
   MtPagination as SwPagination,
   // @deprecated
   MtSkeletonBar as SwSkeletonBar,
+  // @deprecated
+  MtToast as SwToast,
```
Haberkamp, line 100: Making things backward compatible makes sense for the components we already offer. I suggest that we don't offer that for new components for the following reasons.
1. We don't offer stuff we directly deprecate
2. Leaving this out means that people have less to migrate because they instantly use the correct one

packages/component-library/src/components/feedback-indicator/mt-toast/mt-toast.vue
```diff
@@ -0,0 +1,245 @@
...
+  bottom: 0;
+  left: 0;
+  width: 100%;
+  text-align: center;
+
+  &__bottom-left {
+    flex: 1;
+  }
+
+  &__bottom-center {
```
Haberkamp, line 126: Those positions look to me more like modifiers instead of child elements. According do BEM the `__` would need to be swapped out against `--`

Haberkamp (approved):

.changeset/nervous-spoons-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": minor
+---
+
+Implement toasts
```
jleifeld, line 5: I write general foundings from my UI testing here:

- Adding several toasts in the middle shows a animation very shortly which looks a bit weird. If we can fix this this would be cool
- Hover effect for closing "X" is missing (if wanted by design)
- A title attribute for long text would be good so that it could be readable on hover (https://www.w3schools.com/tags/att_global_title.asp#:~:text=HTML%20title%20Attribute&text=The%20title%20attribute%20specifies%20extra,it%20is%20not%20necessarily%20useful).)

jleifeld (changes requested):

packages/component-library/src/components/feedback-indicator/mt-toast/mt-toast.vue
seggewiss: It's not entirely possible to scope all the styling since we change childs of mt-toast-notification. But I extracted that part and made the rest scoped.

.changeset/nervous-spoons-glow.md
seggewiss: Thanks to your help I was able to resolve all of this 😊

jleifeld (approved):

Weltraumakustik (approved):

PR #119, opened 2024-04-05
Flex wrap mt-select-selection-list

atloss:
## What?
The `.mt-select-selection-list` class needs to be flex wrapped.

## Why?
Otherwise the `.mt-select-selection-list__item-holder` elements are overflowing the select if there are too many selected

## Screenshots
Before:
mt-select-without-flex-wrap

After:
mt-select-with-flex-wrap


Haberkamp (changes requested):
Thanks for creating a PR @atloss. Would be great to have a changelog and also a new interactive story. This will create a new screenshot test to ensure your change stays and won't be part of a regression.

atloss:
@Haberkamp Thanks for your feedback! Pardon, if I'm still doing something wrong. I added the interaction test, however I don't know how to correctly test the wrapping, since it's not emitting an event or something similar. I will adjust the changelog once my test makes sense. Do you have any feedback for me considering the test?

Haberkamp:
To adjust the viewport size of that test you can use code below.

```ts
// use smaller viewport to test wrapping of multi select
if (
  context.id ===
  "interaction-tests-form-mt-select--visual-test-ensure-correct-multi-selection-wrapping"
) {
  await page.setViewportSize({ width: 500, height: 650 });
}
  ```
  
  This needs to be added to the `test-runner.ts` file.
  
  After you push the commit you can download the new screenshot from the job summary and commit that as well.

atloss:
@Haberkamp Thank you, I adjusted the `test-runner.ts` file accordingly. However, unfortunately I don't know what you mean with

> [...] you can download the new screenshot from the job summary and commit that as well.

Haberkamp:
Hey, yeah currently the process of updating is a bit complex. Here are the things you need to do:
1. Update the `test:storybook` command of the component library to `test-storybook -u`
2. Commit and push that
3. Click on the "Checks" tab of this PR
4. Open the "Visual tests" item in the left sidebar, (You should see a page that looks like [this](https://github.com/shopware/meteor/actions/runs/8646794190))
5. Scroll down a bit until you see an "Artifacts" section
6. Download the artifacts
7. Put all of the screenshots into the `__snapshots__` folder
8. Remove the `-u` flag you edited in step 1.
9. Commit and push

I tried to do it myself, but I don't have the rights to push to your branch, cause it's a fork.

Haberkamp (approved):

Haberkamp:
Thanks for your contribution, your changes will be part of the next release. 😊

PR #133, opened 2024-04-12
Make create id compatible with webpack

Haberkamp:
## What?

This PR makes the id generation compatible with webpack.

## Why?

The previous solutions made the webpack builds fail. This was because webpack could not resolve the nodejs crypto api in a browser context.

## How?

I replace the crypto api with nanoid. A package for generating ids that it cross-platform compatible.

## Testing?

I tested it on the nuxt example app as well as with a webpack project.


jleifeld (approved):
Looks good 👍

PR #136, opened 2024-04-15
Import popover component to use

nguyenquocdaile:
## What?
The Popover component is already implemented in [storybook](https://meteor-component-library.vercel.app/?path=/docs/components-overlay-mt-popover--docs) but it's not imported to use [here](https://github.com/shopware/meteor/blob/main/packages/component-library/src/index.ts)

### DIFF
--- a/.changeset/many-coins-play.md
+++ b/.changeset/many-coins-play.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": minor
+---
+
+Import overlay popover component to use.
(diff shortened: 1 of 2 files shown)


Haberkamp (changes requested):
Hey, thanks for the PR. Please add a changelog with a minor version bump. This way other devs will be aware of what changed. If you need help with that you can check out the contribution guidelines in the repository.

nguyenquocdaile:
> Hey, thanks for the PR. Please add a changelog with a minor version bump. This way other devs will be aware of what changed. If you need help with that you can check out the contribution guidelines in the repository.

Hi, I added it 😊

Haberkamp (approved):

jleifeld (changes requested):
Would also export popover items. Otherwise the component can't be used correctly

nguyenquocdaile:
> Would also export popover items. Otherwise the component can't be used correctly

Yes, I also export popover-item-result too 😊

jleifeld (approved):

PR #138, opened 2024-04-19
Add a prop to remove settings table button in the header

nguyenquocdaile:
## What?
There are two issues in this PR 

1. If we set two props `disableEdit` and `disableDelete` is true. It will have a small bug UI 

Screenshot 2024-04-19 at 11 15 00

2. I add props to disable the settings table. And remove the last column when set three props `disableEdit`, `disableDelete`, `disableSettingsTable` are true.


Haberkamp:
@nguyenquocdaile hey good catch, one small thing. Instead of adding another prop I suggest that we hide the context menu button if both `disableEdit` and `disableDelete` are set to true trough a computed property.

nguyenquocdaile:
> @nguyenquocdaile hey good catch, one small thing. Instead of adding another prop I suggest that we hide the context menu button if both `disableEdit` and `disableDelete` are set to true trough a computed property.

Hi @Haberkamp, Another prop ( disableSettingsTable ) is to disable settings table here :D. it's not related to the context menu button.

Screenshot 2024-04-22 at 13 57 10



If a set three props `disableEdit`, `disableDelete`, `disableSettingsTable` is true. The last column will remove also.

jleifeld (changes requested):
Looks also alright 🙂 Would be also cool to add a visual test to it to make sure we don't break it in the future

jleifeld (approved):

PR #142, opened 2024-04-23
Allow add an image to cell table

nguyenquocdaile:
## What?
Following the idea in the product list, sometimes we need to add an image to a cell table. On the other hand, in the current app, we also need to implement an image in the cell table.

image


jleifeld (changes requested):
Looks good! Can you also add a visual test for the image rendering?

nguyenquocdaile:
> Looks good! Can you also add a visual test for the image rendering?

hmm. I tried to add a visual test, but it always failed the snapshot test. But it's not fail on my local. It seems the server test doesn't support an image 🤔

PR #143, opened 2024-04-24
PS-224 - Update and build globally mt-empty-state

sydinh:
## What?

Updating and building globally mt-empty-state

## Why?

The mt-empty-state component should be used more widely, not just internally @shopware-ag/meteor-component-library

## How?

Just export it

## Testing?

N/A

## Screenshots (optional)

N/A

## Anything Else?

N/A


packages/component-library/src/components/layout/mt-empty-state/mt-empty-state.vue
```diff
@@ -11,6 +11,14 @@
     <p class="mt-empty-state__description">
       {{ description }}
     </p>
+
+    <div v-if="$slots.link" class="mt-empty-state__link">
```
jleifeld, line 15: I would replace both slots with properties instead. This makes future non-breaking changes easier because they are more strict. And people can't use the slots for unwanted behavior 🙂 

(Just a general tip: `v-if="$slots.link"` is a anti-pattern we discovered in the last years. The v-if didn't get updated when the slot gets filled after the component is mounted. I would prefer removing the v-if and using a CSS ":empty" selector with "display:none". But in this case as mentioned I would replace it with props)

jleifeld (commented):
Good that you included tests, etc. Only have one small thing

packages/component-library/src/components/layout/mt-empty-state/mt-empty-state.vue
jleifeld: FYI: prop naming should match the Figma file: https://www.figma.com/file/Krub3xbG0vMUdoL8vzEZ7l/%E2%98%84%EF%B8%8F-Meteor-Admin-Foundations-%E2%80%93-6.0.0?type=design&node-id=10044-307923&mode=design&t=EkJ5jb3x0gwZ5crD-0

Haberkamp:
Looks good so far, just one small thing. I have the feeling that the icons are too large. I looked it up in Figma and there they're way smaller.

Bildschirmfoto 2024-04-24 um 08 25 34

packages/component-library/src/components/layout/mt-empty-state/mt-empty-state.vue
sydinh: Thanks @jleifeld. There will be pretty many things that I need to change/adjust after your feedback. One question, should I group the related properties (props) into a single object, for example 👇🏻  

```shell
mt-empty-state
    icon
    content
        headline*
        description*
        link
    action
```

Thanks for the tip too. I will discover it one day 😉

sydinh:
> Looks good so far, just one small thing. I have the feeling that the icons are too large. I looked it up in Figma and there they're way smaller.
> 
> Bildschirmfoto 2024-04-24 um 08 25 34

I have the same feeling as you, I basically got it from https://developer.shopware.com/resources/meteor-icon-kit/, but I am unsure why it was so large 😭

packages/component-library/src/components/layout/mt-empty-state/mt-empty-state.vue
jleifeld: We should try to use the same properties as in the figma file: 

Bildschirmfoto 2024-04-24 um 09 42 46

This would makes it easier to react to changes if a designer changes the component in Figma. In some cases it would make sense to modify it a bit. Example: Instead of "Button: boolean" a "Button:string" value would make more sense. Otherwise we would need another prop just for the button content.

Haberkamp:
@sydinh the `mt-icon` component has a `size` prop. You can use that to define the size of the icon. One thing that crossed my mind, could you add tokens to the component? That way we also support dark mode. 🌚 You can toggle the theme via a button in the top toolbar. It has a picture icon.
Bildschirmfoto 2024-04-24 um 14 03 10

packages/component-library/src/components/layout/mt-empty-state/mt-empty-state.vue
sydinh: ✔️

jleifeld (approved):

PR #156, opened 2024-05-02
NEXT-35651 - add tokens to switch component

Haberkamp:
## What?

I added token support for the switch component.

## Why?

This brings the component closer to the designs in Figma.

## How?

I replaced the SCSS variables with the design tokens.

## Testing?

You can check out the changes in the branch preview below.


packages/component-library/src/components/form/mt-switch/mt-switch.vue
```diff
@@ -265,6 +265,7 @@ $mt-field--switch-color-error: $color-crimson-500;
       display: grid;
       grid-template-columns: 24px 1fr auto;
       align-items: center;
+      // TODO: what does this do?
```
jleifeld, line 268: missing todo

jleifeld (changes requested):

packages/component-library/src/components/form/mt-switch/mt-switch.vue
seggewiss: We need the TODO rule everwhere :D

jleifeld (approved):

PR #161, opened 2024-05-07
NEXT-35973 - Allow getting user's timezone

sydinh:
## What?

I'm allowing Shopware apps to get the user's timezone.

## Why?

Some applications need the user's timezone to calculate and display data based on it.

## How?

```ts
const userTimezone = await sw.context.getUserTimezone();
```

## Testing?

N/A

## Screenshots (optional)

N/A

## Anything Else?

N/A

### DIFF
--- a/.changeset/soft-years-hope.md
+++ b/.changeset/soft-years-hope.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-admin-sdk": minor
+---
+
+Allow getting user's timezone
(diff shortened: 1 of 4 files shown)


jleifeld (changes requested):
Looks good! Small thing: can you add a information in the documentation since which Shopware version this feature will be available?

sydinh:
> Looks good! Small thing: can you add a information in the documentation since which Shopware version this feature will be available?

I just left a caution. I probably come back to remove it after I add the corresponding support in Shopware. @jleifeld

jleifeld (approved):

PR #164, opened 2024-05-07
NTR - Updated context.md

sydinh:


seggewiss (approved):

seggewiss:
> Can you replace this with the Shopware version where it is available? So that it doesn't happen if someone tries out the feature with a older Shopware version and then it doesn't work. Similar like this:
> 
> https://github.com/shopware/meteor/blob/177026ad9aab6f638ad5264ead77e4b48b60813b/docs/admin-sdk/docs/guide/2_api-reference/toast.md?plain=1#L4

We don't have this for other features either.

jleifeld (approved):

jleifeld (approved):

PR #167, opened 2024-05-13
NEXT-35659 - add tokens for data table

Haberkamp:
## What?

I added tokens for the data table component.

## Why?

This brings the component closer to the design system.

## How?

I replaced the SCSS variables with CSS variables.

## Testing?

You can checkout the changes in the branch preview below.


packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
```diff
@@ -2729,11 +2723,12 @@ table.is--dragging-inside {
   &.mt-data-table__table-head-add-column-indicator {
     cursor: pointer;
     z-index: 10;
-    background-color: $color-white;
 
     .mt-icon {
       display: block;
       color: $color-darkgray-200;
+      background: #fff;
```
fabianhueske, line 2730:
```suggestion
      color: var(--color-text-secondary-default);
      background: var(--color-icon-inverted-default);
```

Maybe 🤷‍♂️ Depends on where this is applied 😄

packages/component-library/__snapshots__/interaction-tests-table-and-list-mt-data-table--visual-test-add-column-indicator-popover-snap.png
jleifeld, line 1: The contrast in the header is not given to the bar with search and buttons. Is this really wanted? In the screendesigns of the data table this contrast is wanted

jleifeld (changes requested):

jleifeld (approved):

Issue #168, opened 2024-05-16
Add ability to prevent auto close for mt-toast


quando1910:
we have a case showing toast (type critical without any action button) when the app loses the Internet Connection, and it will not close until the Internet is available again.

Could we add the new options to override the default auto-close behavior ?

PR #169, opened 2024-05-16
feat: add auto close options for toast

quando1910:
## What?
Add auto close options for toast
## Why?
Refer: https://github.com/shopware/meteor/issues/168
## How?
create a new option `autoClose`, 
when it true, it will auto-close.
when it false, it will never auto-close.
when it undefine, follow the default behavior in toast matrix.

## Testing?
N/A
## Screenshots (optional)
N/A
## Anything Else?
N/A


seggewiss:
Unfortunately I can't accept this PR. We spend hours upon hours to determine how a toast should behave in the **Shopware Administration**. With that in mind we build the Components which are fed by the `meteor-admin-sdk`.

Here you can find an overview of how toasts need to behave.

If your toast doesn't match that matrix **it's not a toast**! It needs to be a notification.

Issue #176, opened 2024-05-21
Add the mt-link


quando1910:
Do we have the plan to add an `mt-link`, which has `:to` props. if `:to` have value so it will render NuxtLink, otherwise, will render a button.
<img width="279" alt="Screen Shot 2024-05-21 at 8 10 54 PM" src="https://github.com/shopware/meteor/assets/29194019/3a5e66e4-2c8d-4546-bfae-7e2fa8200dac">

PR #177, opened 2024-05-22
Feat: Add mt-link component

dieudtq:
## What?
- Missing a component for the internal link,
 https://github.com/shopware/meteor/issues/176
## Why?
- N/A
## How?
- Add `mt-link` component, which has :to props. if :to have value so it will render `router-link`, otherwise, will render `mt-button`.
## Testing?
- Check out branch
- Run storybook and check the `Components/Form/mt-link`


Haberkamp:
@quando1910 Hey 👋 we already have an `mt-external-link` component. I think it makes more sense to adjust that component instead of creating a new one.

quando1910:
> @quando1910 Hey 👋 we already have an `mt-external-link` component. I think it makes more sense to adjust that component instead of creating a new one.

Hello, this mt-link is used for internal case, it will be the NuxtLink or a button based on the location path value.

Haberkamp:
@quando1910 I don't think it's a good idea that we render a button when the developer does not provide `to` property. A link should always look like a link. In cases where we need a link that looks like a button I suggest that we make the button component polymorphic (`<component :is="MY_COMPONENT"`) meaning you can provide a NuxtLink. I suggest the same for the external link component. Let's rename the component and make it polymorphic.

packages/component-library/src/components/form/mt-link/mt-link.interactive.stories.ts
```diff
@@ -0,0 +1,65 @@
...
+    variant: "primary",
+  },
+};
+
+export const VisualTestRenderLinkSecondary: MtLinkStory = {
+  name: "Link secondary",
+  args: {
+    variant: "secondary",
+  },
+};
```
Haberkamp, line 30: We don't have a secondary link component. We don't need that.

packages/component-library/src/components/form/mt-link/mt-link.interactive.stories.ts
```diff
@@ -0,0 +1,65 @@
...
+    to: "/",
+  },
+};
+
+export const VisualTestRenderLinkPrimary: MtLinkStory = {
+  name: "Link primary",
+  args: {
+    variant: "primary",
+  },
+};
```
Haberkamp, line 23: This story is kind of unnecessary because it's the same as the one above.

packages/component-library/src/components/form/mt-link/mt-link.interactive.stories.ts
```diff
@@ -0,0 +1,65 @@
...
+    variant: "critical",
+  },
+};
+
+export const VisualTestRenderLinkSmall: MtLinkStory = {
+  name: "Link positive",
+  args: {
+    size: "small",
+  },
+};
```
Haberkamp, line 44: In Figma we don't have small or large links we only have one size.

packages/component-library/src/components/form/mt-link/mt-link.interactive.stories.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const VisualTestRenderLinkLarge: MtLinkStory = {
+  name: "Link large",
+  args: {
+    size: "large",
+  },
+};
```
Haberkamp, line 51: Same here, we only have one size for links.

packages/component-library/src/components/form/mt-link/mt-link.vue
```diff
@@ -0,0 +1,171 @@
...
+      type: Boolean,
+      required: false,
+      default: false,
+    },
+    underline: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
+    elementType: {
```
Haberkamp, line 60: I would rename that to `as`

packages/component-library/src/components/form/mt-link/mt-link.vue
```diff
@@ -0,0 +1,171 @@
...
+    disabled: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
+    underline: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
```
Haberkamp, line 59: Do we have cases where we don't want an underline? Refering again to the Figma component. But I could be wrong and we may have many cases where we don't have an underline.

packages/component-library/src/components/form/mt-link/mt-link.vue
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  &.mt-link--large {
+    font-size: $font-size-s;
+    font-weight: $font-weight-medium;
+  }
+
+  &:focus-visible {
+    border-color: var(--color-border-brand-selected);
+    outline: none;
```
Haberkamp, line 120: Please use an outline instead of a border. This is because once the user focuses the component the layout shifts because of the border.

Haberkamp (changes requested):

packages/component-library/src/components/form/mt-link/mt-link.interactive.stories.ts
dieudtq: Thank you, I updated it, please help me take a look

packages/component-library/src/components/form/mt-link/mt-link.vue
dieudtq: Thank you, I updated it, please help me confirm

packages/component-library/src/components/form/mt-link/mt-link.vue
dieudtq: Thank you, I updated it, please help me take a look again

packages/component-library/src/components/form/mt-link/mt-link.vue
dieudtq: I guess there will be some cases, like links by product name to open a details page in a table, or some design ideas where the link will not have an underline. It depends on the design concept. 
However, it's also possible that it's just an edge case, I removed it, users can use it themselves with the style they want if they need to.

packages/component-library/src/components/form/mt-link/mt-link.vue
```diff
@@ -0,0 +1,132 @@
...
+        return;
+      }
+
+      this.$emit("click", event);
+    },
+  },
+});
+</script>
+
+<style lang="scss">
```
Haberkamp, line 88: It makes sense to scope these styles. This way other components cannot mess with the styles. It's still possible to apply margin to the root element so that you're able to position it in case you need that.

Haberkamp (changes requested):

packages/component-library/src/components/form/mt-link/mt-link.interactive.stories.ts
```diff
@@ -0,0 +1,44 @@
+import { within } from "@storybook/test";
+import { expect } from "@storybook/test";
+
+import meta, { type MtLinkMeta, type MtLinkStory } from "./mt-link.stories";
+
+export default {
+  ...meta,
+  title: "Interaction Tests/Form/mt-link",
```
Haberkamp, line 8: It makes more sense to move the both story files unter `Navigaton/mt-link`

packages/component-library/src/components/form/mt-link/mt-link.vue
dieudtq: Thank for the feedback @Haberkamp, I updated it.

packages/component-library/src/components/form/mt-link/mt-link.interactive.stories.ts
dieudtq: Thanks, @Haberkamp , I moved it to `Navigation/mt-link`.

packages/component-library/src/components/form/mt-link/mt-link.interactive.stories.ts
dieudtq: Thank you, I removed it.

Haberkamp (approved):

Issue #207, opened 2024-06-04
Add slots for mt-data-table


quando1910:
We've faced with the issue can not custom a cell of mt-data-table.
For example: with product name column, we want to show the product name and the variant below in the same column and style for it.
<img width="907" alt="Screen Shot 2024-06-04 at 10 25 45 AM" src="https://github.com/shopware/meteor/assets/29194019/5ece5f43-d212-43b4-bf0f-3a684172dd67">

Could we have plan to add slot for each cell of mt-data-table, then we can customize it ?

fabianhueske:
Hi @quando1910 😊 ✌️ Just out of curiosity, for which project do you need this? 🤔 We are definitely thinking of adding a possibility like this but we need to get in all requirements first.

PR #210, opened 2024-06-04
Feat: add slot for mt-data-table

quando1910:
## What?
Refer this issue: https://github.com/shopware/meteor/issues/207
## Why?
To have more chances for mt-data-table customization 
## How?
Just add some slots. Not change any logic.
## Testing?
N/A
## Screenshots (optional)
N/A
## Anything Else?
N/A

### DIFF
--- a/packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
+++ b/packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
@@ -9,40 +9,48 @@
           :model-value="searchValue"
           @change="emitSearchValueChange"
         />
+        <template v-if="filters.length > 0">
+          <slot
+            name="filter-button"
+            v-bind="{ filterChildViews, isOptionSelected, addOption, removeOption }"
+          >
+            <mt-popover title="Filters" :child-views="filterChildViews">
+              <template #trigger="{ toggleFloatingUi }">
+                <mt-button variant="secondary" @click="toggleFloatingUi">
+                  <mt-icon name="solid-filter-s" aria-hidden="true" />
 
-        <mt-popover v-if="filters.length > 0" title="Filters" :child-views="filterChildViews">
-          <template #trigger="{ toggleFloatingUi }">
-            <mt-button variant="secondary" @click="toggleFloatingUi">
-              <mt-icon name="solid-filter-s" aria-hidden="true" />
+                  <span>{{ $t("mt-data-table.filter.addFilter") }}</span>
+                </mt-button>
+              </template>
 
-              <span>{{ $t("mt-data-table.filter.addFilter") }}</span>
-            </mt-button>
-          </template>
+              <template #popover-items__base="{ changeView }">
+                <mt-popover-item
+                  v-for="filter in filters"
+                  :key="filter.id"
+                  :label="filter.label"
+                  show-options
+                  @click-options="() => changeView(filter.id)"
+                />
+              </template>
 
-          <template #popover-items__base="{ changeView }">
-            <mt-popover-item
-              v-for="filter in filters"
-              :key="filter.id"
-              :label="filter.label"
-              show-options
-              @click-options="() => changeView(filter.id)"
-            />
-          </template>
-
-          <template v-for="filter in filters" :key="filter.id" #[`popover-items__${filter.id}`]>
-            <mt-popover-item
-              v-for="option in filter.type.options"
-              :key="option.id"
-              :label="option.label"
-              show-checkbox
-              :checkbox-checked="isOptionSelected(filter.id, option.id)"
-              @change-checkbox="
-                (isChecked) =>
-                  isChecked ? addOption(filter.id, option.id) : removeOption(filter.id, option.id)
-              "
-            />
-          </template>
-        </mt-popover>
+              <template v-for="filter in filters" :key="filter.id" #[`popover-items__${filter.id}`]>
+                <mt-popover-item
+                  v-for="option in filter.type.options"
+                  :key="option.id"
+                  :label="option.label"
+                  show-checkbox
+                  :checkbox-checked="isOptionSelected(filter.id, option.id)"
+                  @change-checkbox="
+                    (isChecked) =>
+                      isChecked
+                        ? addOption(filter.id, option.id)
+                        : removeOption(filter.id, option.id)
+                  "
+                />
+              </template>
+            </mt-popover>
+          </slot>
+        </template>
 
         <slot name="toolbar" />
       </div>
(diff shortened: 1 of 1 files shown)


packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
```diff
@@ -417,34 +425,36 @@
...
 
                     <template v-else>
-                      <!-- Use the correct renderer for the column -->
-                      <mt-data-table-number-renderer
-                        v-if="column.renderer === 'number'"
-                        :data="data"
-                        :column-definition="column"
-                        @click="$emit('open-details', data)"
-                      />
+                      <slot :name="`column-${column.property}`" v-bind="{ data, column }">
```
jleifeld, line 428: I think this can be insecure because this allows also overriding existing renderer. Better would be an approach like this:
```html
<template v-else>
  <!-- Use the correct renderer for the column -->
  <mt-data-table-number-renderer
    v-if="column.renderer === 'number'"
    :data="data"
    :column-definition="column"
    @click="$emit('open-details', data)"
  />

  <mt-data-table-text-renderer
    v-else-if="column.renderer === 'text'"
    :data="data"
    :column-definition="column"
    @click="$emit('open-details', data)"
  />

  <mt-data-table-badge-renderer
    v-else-if="column.renderer === 'badge'"
    :data="data"
    :column-definition="column"
    @click="$emit('open-details', data)"
  />

  <mt-data-table-price-renderer
    v-else-if="column.renderer === 'price'"
    :data="data"
    :column-definition="column"
    @click="$emit('open-details', data)"
  />

  <slot name="custom-column-renderer" :v-bind="{ data, column }"></slot>
</template>
```

Then it is possible to add new, custom renderer which can be defined in the column definition. E.g.
```html
<template #custom-column-renderer>
    <my-custom-renderer v-if="column.renderer ==== 'my-custom-renderer' />
</template>
```

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
```diff
@@ -453,23 +463,27 @@
...
-                    />
-
-                    <mt-context-menu-item
-                      v-if="!disableDelete"
-                      type="critical"
-                      :label="$t('mt-data-table.contextButtons.delete')"
-                      @click="$emit('item-delete', data)"
-                    />
-                  </mt-context-button>
+                  <slot name="action-button" v-bind="{ data }">
```
jleifeld, line 466: Same here. If someone uses this slot some logic with "disableEdit", etc. can get broken by using the slot wrong (e.g. the v-if before does not render the button even if the logic would make sense for it).

I would prefer to implement it in a more safe way with a slot just around the text. Then it is also customizable but is more stable.

```html
<a v-if="!disableEdit" href="#" @click.prevent="$emit('open-details', data)">
  <slot name="action-button" v-bind="{ data }">
    {{ $t("mt-data-table.contextButtons.edit") }}
  </slot>
</a>
```

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
```diff
@@ -453,23 +463,27 @@
...
+                    <a v-if="!disableEdit" href="#" @click.prevent="$emit('open-details', data)">
+                      {{ $t("mt-data-table.contextButtons.edit") }}
+                    </a>
+                  </slot>
+                  <slot name="context-button" v-bind="{ data }">
```
jleifeld, line 471: Same here. This would allows to override the context buttons at a whole. I would prefer a property instead of a slot here. Something like this:
```js
additionalContextButtons: {
    type: Array as PropType<Array<{
       // Here the needed information about the buttons
    }>>,
    required: false,
   default: undefined
}
```

Then we can iterate over this property in the template and can also adjust the `v-if` check for the context button:
```
<mt-context-button v-if="!(disableDelete && disableEdit) || additionalContextButtons?.length > 0">
```

And then do a v-for for all context buttons in the property

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
```diff
@@ -9,40 +9,48 @@
           :model-value="searchValue"
           @change="emitSearchValueChange"
         />
+        <template v-if="filters.length > 0">
```
jleifeld, line 12: @Haberkamp I think this implementation could be unstable very quickly. You are more into the topic of filters. Do you have a better, more defined approach for adding custom filter types?

I could imagine a general slot which allows adding custom "filter renderer"

jleifeld (changes requested):

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
quando1910: I think we can add the v-else-if for slot also. so we can control all the current content.
```
<template v-else>
  <!-- Use the correct renderer for the column -->
  <mt-data-table-number-renderer
    v-if="column.renderer === 'number'"
    :data="data"
    :column-definition="column"
    @click="$emit('open-details', data)"
  />

  <mt-data-table-text-renderer
    v-else-if="column.renderer === 'text'"
    :data="data"
    :column-definition="column"
    @click="$emit('open-details', data)"
  />

  <mt-data-table-badge-renderer
    v-else-if="column.renderer === 'badge'"
    :data="data"
    :column-definition="column"
    @click="$emit('open-details', data)"
  />

  <mt-data-table-price-renderer
    v-else-if="column.renderer === 'price'"
    :data="data"
    :column-definition="column"
    @click="$emit('open-details', data)"
  />

  <slot v-else-if="column.renderer === 'custom'" name="custom-column-renderer" :v-bind="{ data, column }"></slot>
</template>
```

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
quando1910: by this way, we can not change the action for this button, for example, I want to replace by a "view details" button instead of the edit button.

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
quando1910: should we use v-else for slot ?
```js
<slot v-else name="custom-column-renderer" v-bind="{ data, column }" />
```

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
quando1910: FYI, the current behavior is not proper with tablet, because it's open a popover, we want to customize the filter button, it will open the modal instead.

packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.vue
quando1910: I updated, please check again

seggewiss:
Sorry I just stumbled again across this. Could you rebase it and see that the Check pass? 😊

seggewiss:
There was no activity on this PR for more than a month now, I'll close it. Feel free to reopen anytime.

PR #223, opened 2024-06-17
NEXT-36816 - Add modal component

Haberkamp:
## What?

I added a modal component.

## Why?

Some projects already require a modal component, so they built their own. Sharing a single modal component improves consistency. The new modal component is also more accessible as it has better screen-reader support and supports keyboard navigation.

## How?

I added a new modal component

## Testing?

You can checkout the changes in the branch preview below. You can also see an example in the nuxt example application.

## Anything Else?

* Added an `mt-inset` component that allows insetting components using a more composable approach
* This pr introduces testing-library. This way we can also test for a11y issues during testing, moreover it reduces the number of false positives.


packages/component-library/src/components/overlay/mt-modal/mt-modal.vue
```diff
@@ -0,0 +1,205 @@
...
+
+.mt-modal__content {
+  --mt-modal-padding: 1.5rem;
+  --mt-modal-padding-bottom: 2.5rem;
+  --inset-top: var(--mt-modal-padding);
+  --inset-right: var(--mt-modal-padding);
+  --inset-bottom: var(--mt-modal-padding-bottom);
+  --inset-left: var(--mt-modal-padding);
+
+  padding: var(--mt-modal-padding);
```
fabianhueske, line 153:
```suggestion
  padding: 1.5rem;
```

Let's simplify (and optically fix) this but having all content paddings the same 😊

packages/component-library/src/components/overlay/mt-modal/mt-modal.vue
```diff
@@ -0,0 +1,205 @@
...
+  align-items: baseline;
+  column-gap: 0.5rem;
+}
+
+.mt-modal__close-button {
+  cursor: pointer;
+  color: var(--color-icon-primary-default);
+  border-radius: $border-radius-default;
+  width: 2rem;
+  height: 2rem;
```
fabianhueske, line 191:
```suggestion
  width: 2.5rem;
  height: 2.5rem;
```
Let's make this button a bit bigger to align to the recommended 40px minimum touch target size. Btw. 2.5rem only works because our default font-size (rem) seems to be 16px? Shouldn't it be 14px or do we keep them intentionally out of sync to have a rem size that's fitting in the 8px grid?

Haberkamp: Yes, I would leave it like this. If the token for the default font size will 14px it will be 0.875rem. The target size for WCAG AAA is 44px, it would then set it to 44px.

packages/component-library/src/components/layout/mt-inset/mt-inset.vue
```diff
@@ -0,0 +1,12 @@
+<template>
```
jleifeld, line 1: Would it make more sense to set the `inset` mode via a prop?

packages/component-library/src/components/overlay/mt-modal/sub-components/mt-modal-root.vue
```diff
@@ -0,0 +1,63 @@
+<template>
```
jleifeld, line 1: Just for interest: what is the benefit of having separate components for root, trigger, inset, etc. instead of one modal component with `isOpen` property?

packages/component-library/src/components/layout/mt-inset/mt-inset.vue
Haberkamp: Yes, it does. I will do that.

packages/component-library/src/components/overlay/mt-modal/sub-components/mt-modal-root.vue
Haberkamp: The idea was to provide simpler developer experience. You still are able to use an isOpen prop in cases where you need that. For example when the modal opens instantly when the page loads or when you want to programatically show the modal let's say you want to show it 30 minutes after the user has been inactive. But for the 99% of cases a modal gets opened when the user clicks on a button that looks like a link or a normal button and the user and a modal will almost always have an cancel and a confirm button. Those are just tiny abstractions that won't add up to much bundle size. If someones does not prefer to use them they can make use of the `isOpen` prop and the other components will be tree-shaked. 

An example: 

```vue
<mt-modal-root>
  <mt-modal>
    <template #default>
      <p>My content</p>
    </template>
    
    <template #footer>
       <mt-modal-close :as="MtButton" variant="secondary">Close</mt-modal-close>
    
       <mt-modal-action 
         :as="MtButton"
         variant="primary"
         @click="async (close) => {
           await makeRequest();
         
           close();
         }"
         >Confirm</mt-modal-close>
    </template>
  <mt-modal>
</mt-modal-root>
```

There's no need to define an isOpen property, if not wanted.

Opposed to the work a developer has to perform in platform right now:

```vue
<template>
  <sw-modal v-if="isOpen">
    <template #default>
      <p>My content</p>
    </template>

    <template #footer>
      <sw-button>Close</sw-button>

      <sw-button
        @click="await () => {
          async makeRequest();
          isOpen = false;
         }"
      >
        Confirm
      </sw-button>
    </template>
  </sw-modal>
</template>

<script lang="ts">
export default {
  data() {
    return {
      isOpen: false,
    };
  },
};
</script>
```

I know the difference is not much, but I still believe that doing many little improvements in developer experience add up quickly.

seggewiss (approved):

PR #227, opened 2024-06-26
NEXT-36740: In-App Purchase Admin SDK trigger

Bird87ZA:
## What?
Added In-App Purchases skeleton to AdminSDK. Currently it only needs to open a non-customisable modal.

## Why?
Part of our new In-App Purchases epic

## Testing?
Create a test app, and add a button following the documentation

## Screenshots (optional)
image
image


packages/admin-sdk/src/in-app-purchases/index.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Trigger the in-app purchase flow modal.
+ */
+export const trigger = createSender('inAppPurchases');
+
+export type inAppPurchases = {
+    responseType: unknown,
+
+    featureName: string,
+
+    priceModel: string,
```
mstegmeyer, line 13: aren't we handling this now for the merchant? The selection of the price model? Then we don't need this. May we add it later as an optional string (though actual it's an integer id in the SBP).

cyl3x (approved):

mstegmeyer (approved):

cyl3x (approved):

docs/admin-sdk/docs/guide/2_api-reference/in-app-purchases/in-app-purchases.md
```diff
@@ -0,0 +1,15 @@
+# In-App Purchase Flow
```
jleifeld, line 1: Looks good. One small thing I would add is the first Shopware Version where this feature is available at the top, see https://developer.shopware.com/resources/admin-extension-sdk/api-reference/cms/registerCmsBlock.html as an example

jleifeld (changes requested):

docs/admin-sdk/docs/guide/2_api-reference/in-app-purchases/in-app-purchases.md
mstegmeyer: Maybe we can add this later, since this is only going into a feature-branch? Right now, we have no clue, what the release minor will be, still a lot of dependency on the SBP.

Bird87ZA: I'm adding it anyway as 6.7.0, just so we have it. We can always change it.

jleifeld (approved):

PR #241, opened 2024-07-17
NEXT-37297 - add mt-text component

Haberkamp:
## What?

Adds a `mt-text` component and migrates all font variables over to tokens.

## Why?

Before the solution you had to define every property like font size, line height, weight and family separate. With the text component you just define the size and the color and you're good to go.

## How?

Added a `mt-text` component where you can specify what color and size you want.

## Testing?

The visual tests should be enough testing. But you can still do some manual testing if you want to.

## Anything Else?

There are a handful of cases where we could had code every property instead of using the text component.


packages/component-library/__snapshots__/interaction-tests-form-mt-datepicker--visual-test-datepicker-should-open-snap.png
fabianhueske, line 1: Seems like the clear icon is not vertically centered 👀

packages/component-library/__snapshots__/interaction-tests-form-mt-select--visual-test-disabled-snap.png
fabianhueske, line 1: Not sure if this text isn't to big 🤔 Maybe wrong token? Should be xs

packages/component-library/__snapshots__/interaction-tests-form-mt-select--visual-test-multi-select-snap.png
fabianhueske, line 1: Alignment issues 🙈

packages/component-library/__snapshots__/interaction-tests-layout-mt-card--visual-test-extended-page-snap.png
fabianhueske, line 1: I guess the "MM" icon slot should be center aligned vertically or the subtitle should be smaller like 2xs right? Does not seem to be aligned right now.

packages/component-library/__snapshots__/interaction-tests-navigation-mt-segmented-control--visual-test-render-segmented-control-with-popover-second-level-snap.png
fabianhueske, line 1: Heading alignment broken

fabianhueske (approved):

PR #247, opened 2024-07-29
fix: fully implement x-small mt-button size

bojanrajh:
## What?

Allows `x-small` size for  `mt-button`.

## Why?

Validation for `size` prop failed, but the rest of CSS is there.

## How?

By allowing `x-small` and adding a story.

## Testing?

Storybook.


Haberkamp (changes requested):
Please add a changelog and update the visual tests

.changeset/new-buttons-join.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": patch
+---
+
+Test coverage for x-small mt-button
```
Haberkamp, line 5: I would re-write this to something that more user-facing like add x-small variant to button component or allow x-small as valid variant for button

bojanrajh, line 5:
```suggestion
allow x-small as valid variant for button
```

Haberkamp (approved):

PR #248, opened 2024-07-30
feat: re-export MtSearch in dist

bojanrajh:
## What?

Re-export `MtSearch` component.

## Why?

So we (and everyone else) can re-use it.

### DIFF
--- a/.changeset/clever-swans-talk.md
+++ b/.changeset/clever-swans-talk.md
@@ -1,0 +1,6 @@
+---
+"@shopware-ag/meteor-component-library": patch
+"nuxt-app": patch
+---
+
+Re-export mt-search and reuse it in nuxt-app example
--- a/examples/nuxt-app/app.vue
+++ b/examples/nuxt-app/app.vue
@@ -20,6 +20,8 @@
         </MtText>
 
         <SwTextField label="A text field" v-model="text" />
+
+        <MtSearch />
       </SwBlockStack>
 
       <MtModalRoot>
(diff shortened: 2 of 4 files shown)


Haberkamp (changes requested):
Please add a changelog for this

Haberkamp (approved):

examples/nuxt-app/app.vue
```diff
@@ -20,6 +20,8 @@
         </MtText>
 
         <SwTextField label="A text field" v-model="text" />
+
+        <MtSearch />
```
Haberkamp, line 24: I think just making the `mt-search` component public is enough. We don't need to use every single component in the nuxt example app.

.changeset/clever-swans-talk.md
```diff
@@ -0,0 +1,6 @@
+---
+"@shopware-ag/meteor-component-library": patch
+"nuxt-app": patch
+---
+
+Re-export mt-search and reuse it in nuxt-app example
```
Haberkamp, line 6: Please only include what's relevant for users of the component library. They just care about the `mt-search` component but not wether the component is used inside the nuxt example app.

Haberkamp (changes requested):

examples/nuxt-app/app.vue
bojanrajh: `mt-search` does depend on the `$t` which is provided [here](https://github.com/shopware/meteor/pull/248/files#diff-12ee38794478c53d0ef713c4a4c11b9587d3afe5f9cc1f8216d63bf5a01da0ac). Though this isn't correct solution (using correct Vue/Nuxt i18n library would be the right approach), we do have an "example app" where we can show that the component actually depends on the i18n library.

bojanrajh:
Closing in favour of https://github.com/shopware/meteor/pull/475

PR #249, opened 2024-07-30
fix: add support for SSR in DeviceHelperPlugin

bojanrajh:
## What?

This PR updates nuxt-app example with `MtTabs` component which is using `DeviceHelper` plugin. 

## Why?

To allow server-side rendering when `DeviceHelper` is used.

## How?

By checking for the `window` variable.

## Testing?

Locally, with provided nuxt-app example.


examples/nuxt-app/app.vue
```diff
@@ -20,6 +20,8 @@
         </p>
 
         <SwTextField label="A text field" v-model="text" />
+
+        <MtTabs :items="[{ label: 'Foo' }, { label: 'Bar' }]" />
```
Haberkamp, line 24: Though this is enough to test that the DeviceHelper is SSR compatible, I suggest to make this component do something like switching between content. This way we can be sure that switching views also works when using SSR. Additionally it creates a nicer demo page.

Haberkamp (changes requested):
Please add a changelog

PR #250, opened 2024-07-30
fix: single-option mt-select

bojanrajh:
## What?

This PR sets 100% width for single-option selects.

## Why?

So long-titled options are always displayed.

## How?

By not wrapping flex items and allowing them (it) to be fully expanded. Might be also applied to multi-option selects (because names can differ in the hidden part at the end).


Haberkamp (changes requested):
Please add a changelog here

Haberkamp (approved):

Lucsy3012 (approved):

PR #251, opened 2024-07-31
fix: empty label in base field component

bojanrajh:
## What?

Empty label was taking space, the behaviour is now fixed with CSS solution, as discussed on Slack.

## Why?

So we can align elements properly.

## How?

Hiding empty elements using `:empty` and `:has` selectors. An alternative (and IMO better solution) would be to use `useEmptySlotCheck` as we cannot use CSS to fix all such cases.


Haberkamp (changes requested):
Please add a changelog, you can see how it's done by opening the CONTRIBUTING.md file.

Haberkamp (changes requested):
Please update the visual snapshots to make the pipeline pass

Haberkamp (approved):

PR #252, opened 2024-07-31
fix: disable next and last page button in mt-pagination

nguyenquocdaile:
## What?
We can click the "Next Page" and "Last Page" buttons even when the data table is empty.
https://meteor-component-library.vercel.app/?path=/story/interaction-tests-table-and-list-mt-data-table--visual-test-render-empty-state

## Why?
Calculate the value of totalPages will be zero if totalItems is zero, which will lead to the value of isOnLastPage being false.

## How?
I add a max value as one in the case totalItems is zero.

## Testing?
Unit test

### DIFF
--- a/packages/component-library/src/components/table-and-list/mt-pagination/mt-pagination.spec.ts
+++ b/packages/component-library/src/components/table-and-list/mt-pagination/mt-pagination.spec.ts
@@ -1,5 +1,11 @@
 import { mount } from "@vue/test-utils";
 import MtPagination from "./mt-pagination.vue";
+import { render, screen } from "@testing-library/vue";
+
+// Mock the $t function
+const mockT = (key: any) => {
+  return key;
+};
 
 function createWrapper() {
   return mount(MtPagination, {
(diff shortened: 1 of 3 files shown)


packages/component-library/src/components/table-and-list/mt-pagination/mt-pagination.spec.ts
```diff
@@ -115,5 +115,18 @@ describe("mt-pagination", () => {
 
       expect(wrapper.emitted()["change-current-page"][0]).toStrictEqual([7]);
     });
+
+    it("should disable the next page and last page button when total page value is zero", async () => {
```
Haberkamp, line 119: Please write the test using `@testing-library/vue`. You can see how it's done by opening the `mt-modal.spec.ts` file.

Haberkamp (changes requested):

packages/component-library/src/components/table-and-list/mt-pagination/mt-pagination.spec.ts
```diff
@@ -115,5 +121,29 @@ describe("mt-pagination", () => {
...
+            $t: mockT,
+          },
+        },
+      });
+
+      const nextPageButton = screen.getByTestId("mt-pagination-next-page-button");
+      const lastPageButton = screen.getByTestId("mt-pagination-last-page-button");
+
+      expect(nextPageButton).toBeInTheDocument();
+      expect(lastPageButton).toBeInTheDocument();
```
Haberkamp, line 143: Those assertions are not needed because toBeDisabled requires elements to be inside the DOM so those assertions are redudant.

packages/component-library/src/components/table-and-list/mt-pagination/mt-pagination.spec.ts
```diff
@@ -115,5 +121,29 @@ describe("mt-pagination", () => {
...
+        },
+        global: {
+          mocks: {
+            $t: mockT,
+          },
+        },
+      });
+
+      const nextPageButton = screen.getByTestId("mt-pagination-next-page-button");
+      const lastPageButton = screen.getByTestId("mt-pagination-last-page-button");
```
Haberkamp, line 140: Please retrieve the elements with `[getByRole](https://testing-library.com/docs/queries/byrole/)` method. This makes sure that the components is accessible and actually uses buttons instead of divs to make the component interactive.

Haberkamp (changes requested):

packages/component-library/src/components/table-and-list/mt-pagination/mt-pagination.spec.ts
nguyenquocdaile: I updated the test now. However, the current button does not have a specific role. So Please check if my approach is correct.

Haberkamp (approved):

PR #256, opened 2024-08-14
fix: console warnings

bojanrajh:
## What?

Fixes warning in console `[Vue warn]: Extraneous non-emits event listeners (newItemActive) were passed to component but could not be automatically inherited because component renders fragment or text root nodes. If the listener is intended to be a component custom event listener only, declare it using the "emits" option.` and eslint warning `'defineEmits' is defined but never used`.

### DIFF
--- a/.changeset/clever-clocks-pull.md
+++ b/.changeset/clever-clocks-pull.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": patch
+---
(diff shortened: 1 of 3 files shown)


Haberkamp (changes requested):
Please format the code so prettier will pass and the pipeline turns green

Haberkamp (approved):

PR #261, opened 2024-08-26
NEXT-37979 - Fix  input arrow up and down issue

tajespasarela:
## What?
This MR fixes the issue in the sw-number-field component when pressing the up or down arrow keys if a new value was typed.

## Why?
Before the fix, when the user types some value in a number field and after pressing the arrow up key or arrow down key, the new value is incorrect.

## How?
The solution was to update the internal value on any key input. This way when the arrow keys are pressed, the actual value is there.

## Testing?

Test cases were added

To manually testing:
1. Go to components-form-mt-number-field--docs
3. Focus in the input and type some number
4. Without moving the focus from the field press arrow up or arrow down.

### DIFF
--- a/.changeset/dirty-rules-live.md
+++ b/.changeset/dirty-rules-live.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": patch
+---
+
+fix the issue in the sw-number-field component when pressing the up or down arrow keys if a new value was typed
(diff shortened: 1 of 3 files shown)


packages/component-library/src/components/form/mt-number-field/mt-number-field.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { mount } from "@vue/test-utils";
```
seggewiss, line 1: Instead of creating a new Test file, please write a story for this: 
https://github.com/shopware/meteor/blob/main/packages/component-library/src/components/form/mt-number-field/mt-number-field.interactive.stories.ts

tajespasarela: done

Haberkamp (changes requested):
Please add a changelog using `npx changesets`

Haberkamp (approved):

PR #270, opened 2024-08-30
NEXT-38027 - fix datepicker not emiting value when manually entered

tajespasarela:
## What?
This MR fixes an issue with the mt-datepicker, that provokes not saving the date when it is typed manually instead of selected from the calendar.

## How?

The solution was to emit the changed value also in the `onClose` hook of the `flatpicker` component. This way when user clicks save, it will be emitted.

There is another hook `onValueUpdate` but this is emitted multiple times when typed and it is not emitted when the date is deleted manually. So the `onClose` hook works better.

It was also necessary to update the `flatpickr` dependency to latest version to fix some inconsistent behaviour.

## Testing?

- go to `Should emit date value when manually typed` in storybook


seggewiss (approved):

Haberkamp (changes requested):
Please add a changelog by using `npx changesets`

tajespasarela:
> Please add a changelog by using `npx changesets`

dobe

Haberkamp (approved):

PR #373, opened 2024-11-14
Remove mt external link

alastair-simon:
## What?
Removed deprecated mt-external-link

## Why?
The mt-external-link component has been deprecated in favour of other components

## How?
Removed mt-external-link directory and references


packages/component-library/src/components/layout/mt-empty-state/mt-empty-state.vue
```diff
@@ -48,24 +54,35 @@ defineEmits(["button-click"]);
 </script>
 
 <style scoped>
+.mt-empty-state {
+  display: grid;
```
Haberkamp, line 58: I guess flex with column direction and aligning it at the start is easier. Then you don't need to say `justify-self: start` multiple times

Haberkamp (changes requested):

Lucsy3012 (approved):

PR #377, opened 2024-11-21
NEXT-38863 - implement new text editor component

jleifeld:
## What?

Implement a text editor component with tiptap. Complete documentation and usage of the documentation can be found in Storybook: "Components/Form/mt-text-editor/docs". I recommend to read the documentation there to understand how it can be used.

## Images

Bildschirmfoto 2024-12-11 um 09 33 53
Bildschirmfoto 2024-12-11 um 09 34 09


packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button.vue
```diff
@@ -0,0 +1,97 @@
...
+    "is-active": button.isActive && button.isActive(props.editor),
+  };
+};
+
+const translatedLabel = computed(() => {
+  if (te(props.button.label)) {
+    return t(props.button.label);
+  }
+
+  return props.button.label;
```
Haberkamp, line 64: I don't think the `te` method is needed. If it can't resolve `path.to.translation` it returns the literal path to the token. See this [test](https://github.com/shopware/meteor/blob/main/packages/component-library/src/composables/useI18n.spec.ts#L20)

Haberkamp:
> @Haberkamp This PR uses the native vue-i18n plugin. Can we still merge it or should we wait until you changed all translations to the native vue-i18n plugin?

If it works we don't need to wait until I've changed everything back to `vue-i18n`.  @jleifeld

packages/component-library/.storybook/preview.ts
```diff
@@ -54,6 +55,12 @@ const preview: Preview = {
         { name: "dark", value: DARK_THEME_BACKGROUND_VALUE },
       ],
     },
+    options: {
+      storySort: (a, b) => {
+        // Sort the stories by their id alphabetically
+        return a.id === b.id ? 0 : a.id.localeCompare(b.id, undefined, { numeric: true });
+      },
+    },
```
Haberkamp, line 63: Is this adding much benefit? I guess the default behaviour is that it has the same order as the stories in the code. I personally prefer that because the third story in the sidebar is also the third story in the file

packages/component-library/src/components/form/mt-button/mt-button.vue
```diff
@@ -289,6 +289,7 @@ const isInsideTooltip = useIsInsideTooltip();
   width: 100%;
   display: flex;
   justify-content: center;
+  align-items: center;
```
Haberkamp, line 292: Just curious what problem was this solving? Was it that for block buttons they were not centred?

packages/component-library/src/components/form/mt-colorpicker/mt-colorpicker.interactive.stories.ts
```diff
@@ -2,6 +2,7 @@ import { within, userEvent, fireEvent } from "@storybook/test";
 import { expect } from "@storybook/test";
 
 import meta, { type MtColorpickerMeta, type MtColorpickerStory } from "./mt-colorpicker.stories";
+import { waitUntil } from "@/_internal/test-helper";
```
Haberkamp, line 5: There's a `waitFor` function exported by `@storybook/test` could you use that over this function?

packages/component-library/src/components/form/mt-colorpicker/mt-colorpicker.interactive.stories.ts
```diff
@@ -56,6 +57,188 @@ export const VisualTestOpenColorpicker: MtColorpickerStory = {
...
+    expect(alphaInput).toBeDefined();
+    expect(alphaInput.value).toEqual("81");
+
+    // Check for apply button
+    const applyButton = popover.getByLabelText("colorpicker-apply-color") as HTMLButtonElement;
+    expect(applyButton).toBeDefined();
+  },
+};
+
+export const TestOpenColorpickerWithApplyMode: MtColorpickerStory = {
```
Haberkamp, line 109: We have a `defineStory` function, it's just a small wrapper so you don't need to type `const canvas = createCanvas(canvasElement)` any more. You can just destructure `{ canvas }` from the play function

packages/component-library/src/components/form/mt-colorpicker/mt-colorpicker.interactive.stories.ts
```diff
@@ -56,6 +57,188 @@ export const VisualTestOpenColorpicker: MtColorpickerStory = {
...
+  },
+  play: async ({ canvasElement, args }) => {
+    const canvas = within(canvasElement);
+
+    const pickerToggle = canvas.getByLabelText("colorpicker-toggle");
+
+    await userEvent.click(pickerToggle);
+
+    // Look inside the popover
+    const popover = within(
```
Haberkamp, line 124: When using the `defineStory` function you also don't need this as the scope refers to `document.body`. It than has access to every element inside the iFrame.

packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-color.vue
```diff
@@ -0,0 +1,85 @@
...
+
+<script lang="ts">
+export const colorButton: CustomButton = {
+  name: "text-color",
+  label: "mt-text-editor-toolbar-button-color.label",
+  position: 2000,
+};
+</script>
+
+<style lang="scss">
```
Haberkamp, line 69: Would be fantastic if this uses scoped styles and also plain css as we want to get rid of SCSS.

packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-link.vue
```diff
@@ -0,0 +1,140 @@
+<template>
+  <mt-text-editor-toolbar-button
```
Haberkamp, line 2: When this button is the only thing that's triggering the modal you can use the `<mt-modal-trigger :as="MtTextEditorToolbarButton">` component. That way you don't need to use the `isOpen` property

You can take a look at the `mt-modal.spec.ts` file to see how that works.

packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-link.vue
```diff
@@ -0,0 +1,140 @@
...
+                linkTarget = checked ? '_blank' : '';
+              }
+            "
+            :aria-label="t('mt-text-editor-toolbar-button-link.openInNewTab')"
+          />
+        </div>
+      </template>
+      <template #footer>
+        <div class="mt-text-editor__link-modal-footer">
+          <mt-button variant="secondary" @click="() => (showLinkModal = false)">
```
Haberkamp, line 34: You can use the `mt-modal-close` component, just like you would use the `mt-modal-trigger` component to close the modal

packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-link.vue
```diff
@@ -0,0 +1,140 @@
...
+<script lang="ts">
+export const linkButton: CustomButton = {
+  name: "link",
+  label: "mt-text-editor-toolbar-button-link.label",
+  icon: "regular-link-xs",
+  position: 12000,
+};
+</script>
+
+<style lang="scss" scoped>
```
Haberkamp, line 134: Please only use CSS

seggewiss (approved):
Nice one! <3

packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-table.vue
```diff
@@ -0,0 +1,221 @@
...
+        label: "Remove table",
+        icon: "regular-times-xs",
+        action: (editor) => editor.chain().focus().deleteTable().run(),
+      },
+    ];
+  },
+};
+</script>
+
+<style scoped lang="scss">
```
Haberkamp, line 215: Please only use CSS

packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button.vue
```diff
@@ -0,0 +1,91 @@
...
+    "is-active": button.isActive && button.isActive(props.editor),
+  };
+};
+
+const translatedLabel = computed(() => {
+  return t(props.button.label);
+});
+</script>
+
+<style>
```
Haberkamp, line 62: Please use scoped styles

packages/component-library/src/composables/useI18n.ts
```diff
@@ -82,5 +82,11 @@ export function useI18n({ messages }: Options) {
...
   }
 
-  return { t: translate };
+  function translationExists(path: string): boolean {
+    return (
+      !!get(messages, `${i18n.locale}.${path}`) || !!get(messages, `${i18n.defaultLocale}.${path}`)
+    );
+  }
+
+  return { t: translate, te: translationExists };
```
Haberkamp, line 91: I guess we no longer need this.

packages/component-library/src/composables/useI18n.ts
```diff
@@ -35,12 +35,12 @@ function limit(value: number, max: number) {
...
 
 export function provideI18n(locale: string = "en") {
-  const state = {
+  const currentInjectionState = {
     ...defaultI18nState,
     locale: locale,
   };
 
-  provide(i18nInjectionKey, state);
+  provide(i18nInjectionKey, currentInjectionState);
```
Haberkamp, line 43: Same of this

packages/component-library/src/components/overlay/mt-modal/sub-components/mt-modal-root.vue
```diff
@@ -26,6 +26,13 @@ watch(isOpen, () => {
   emit("change", isOpen.value);
 });
 
+watch(
+  () => props.isOpen,
+  (value) => {
+    isOpen.value = value;
+  },
+);
+
```
Haberkamp, line 35: Can you write a unit test for this?

packages/component-library/src/components/overlay/mt-modal/mt-modal.interactive.stories.ts
```diff
@@ -24,6 +25,26 @@ export const VisualTestContent = {
   }),
 };
 
+export const TestModalOpening = {
```
Haberkamp, line 28: Can you replace that with a unit test? We decided to opt for unit test for everything that is logic related. They run faster and are cheaper than running playwright. For visual related things we have a handful of visual tests and for cases where we rely on things like the `ResizeObserver` we create Storybook tests.

Haberkamp (changes requested):

packages/component-library/.storybook/preview.ts
jleifeld: I thought it was easier to find the correct story when you have a alphabetical order

packages/component-library/src/components/form/mt-button/mt-button.vue
jleifeld: In some situation a small block button has not vertical centered text (depending on the outer element, e.g. in the colorpicker as the apply button). This was needed to fix it

packages/component-library/src/components/form/mt-colorpicker/mt-colorpicker.interactive.stories.ts
jleifeld: Don't know it yet but sounds good 👍

packages/component-library/src/components/form/mt-colorpicker/mt-colorpicker.interactive.stories.ts
jleifeld: Don't know it yet but will change it then

packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-link.vue
jleifeld: This is cool. I think we need some documentation for the new mt-modal component and how it is working 😄

packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-link.vue
jleifeld: Same as above: I will change it. Documentation for this would be also cool :)

packages/component-library/src/composables/useI18n.ts
jleifeld: You're correct, I will remove it

packages/component-library/src/components/overlay/mt-modal/sub-components/mt-modal-root.vue
jleifeld: I will do it

packages/component-library/src/components/overlay/mt-modal/mt-modal.interactive.stories.ts
jleifeld: Makes sense, I will do it 👍

packages/component-library/src/components/form/mt-colorpicker/mt-colorpicker.interactive.stories.ts
jleifeld: I just looked at the `waitFor` method. Their implementation excepts a error everytime when it should get retried. This would make the code much longer. Here is a comparison:
```js
    await waitUntil(() => args.updateModelValue.mock.calls.length > 0);
    // vs
    await waitFor(() => {
      if (args.updateModelValue.mock.calls.length > 0) {
        return;
      }

      throw new Error("Model value was not updated");
    });
  ```
  
  So I think the `waitUntil` method is much cleaner and shorter

File: packages/component-library/src/components/form/mt-colorpicker/mt-colorpicker.interactive.stories.ts (hunk 6 above)
Comment (jleifeld) [reply packages/component-library/src/components/form/mt-colorpicker/mt-colorpicker.interactive.stories.ts:124]: Nice

File: packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-link.vue (hunk 8 above)
Comment (jleifeld) [reply packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-link.vue:2]: It seems like that it only opens it and I can't do anything besides this (like setting the attributes from the editor, etc.). So it won't work for me.

```
const openLinkModal = () => {
  // Get current link from selection
  linkHref.value = props.editor.getAttributes("link").href ?? "";
  linkTarget.value = props.editor.getAttributes("link").target ?? "";

  showLinkModal.value = true;
};
```

Comment (Haberkamp) [reply packages/component-library/src/components/form/mt-text-editor/_internal/mt-text-editor-toolbar-button-link.vue:2]: Ok, if that's not working for you then it's fine to do it like this.

File: packages/component-library/src/components/overlay/mt-modal/mt-modal.spec.ts [hunk 17]
```diff
@@ -65,6 +66,36 @@ describe("mt-modal", () => {
...
+  <mt-modal title='title'>mt-modal works!</mt-modal>
+</mt-modal-root>`,
+    });
+
+    expect(screen.queryByRole("dialog")).not.toBeInTheDocument();
+
+    // WHEN
+    await fireEvent.click(screen.getByRole("button"));
+
+    console.log(result.container.innerHTML);
```
Haberkamp, line 91: There's console.log in here

packages/component-library/.storybook/preview.ts
Haberkamp: Ok, I would just use the default behaviour of Storybook. You can still use search function if you can't find something.

Haberkamp (changes requested):

jleifeld:
I just saw during the platform migration that the following features are missing:
- placeholder
- error
- general disabled

packages/component-library/.storybook/preview.ts
jleifeld: @Haberkamp Just FYI:
Bildschirmfoto 2024-12-20 um 08 49 02

This is how it looks without sorting. So it doesn't match the order correctly of the folder

packages/component-library/.storybook/preview.ts
```diff
@@ -18,6 +18,7 @@ import "@shopware-ag/meteor-tokens/administration/dark.css";
 
 const i18n = createI18n({
   // something vue-i18n options here ...
+  legacy: false,
```
seggewiss, line 21: We should probably communitcate this change. Eventhough it will be released with the next major all the changes regarding Vue i18n need to be communicated. @Haberkamp @jleifeld

seggewiss (dismissed):

packages/component-library/.storybook/preview.ts
jleifeld: Good thing. I will add it to the changelog!

seggewiss (approved):

Haberkamp (approved):

tajespasarela (approved):

PR #400, opened 2024-11-29
Change default size of button to default

Haberkamp:
## What?

Changes the default size to `default` for the `mt-button` component.

## Why?

When there's size called `default` then it should be the default size.

## How?

I changed the size to `default`

## Testing?

The visual tests will catch issues.

## Anything Else?

Depends on #399

### DIFF
--- a/packages/component-library/src/components/form/mt-button/mt-button.vue
+++ b/packages/component-library/src/components/form/mt-button/mt-button.vue
@@ -52,8 +52,7 @@ const props = withDefaults(
   {
     // @ts-expect-error -- This is to keep the default behaviour
     variant: "",
-    // TODO: Update default value to "default"
-    default: "small",
+    size: "default",
   },
 );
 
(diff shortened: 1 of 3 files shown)


packages/component-library/src/components/form/mt-button/mt-button.vue
```diff
@@ -21,133 +21,60 @@
...
+    size?: "x-small" | "small" | "default" | "large";
+    square?: boolean;
+    block?: boolean;
+    link?: string;
+    isLoading?: boolean;
+  }>(),
+  {
+    // @ts-expect-error -- This is to keep the default behaviour
+    variant: "",
+    default: "default",
```
Lucsy3012, line 55: @Haberkamp wouldn't `size` need to be `"default"`? But maybe I'm just not familiar with the `default` property

alastair-simon (approved):

alastair-simon (approved):

PR #405, opened 2024-11-29
NEXT-17726 - right align numbers in numbers field

jleifeld:
## What?

Right align the numbers in the number field

// EDIT:
Now the field can be toggled to be aligned to the end. In default it is aligned at the start

## Screenshots (optional)

Bildschirmfoto 2024-11-29 um 14 24 46


jleifeld:
@Haberkamp do you know why the Visual Testing pipeline is not red?

fabianhueske:
Hi @jleifeld 😊 I would suggest to leave the default to left aligned and add a prop to switch to right alignment since right aligning number fields is perfect for some cases but only for same. We do have a lot more cases where left alignment fits better.

jleifeld:
> Hi @jleifeld 😊 I would suggest to leave the default to left aligned and add a prop to switch to right alignment since right aligning number fields is perfect for some cases but only for some. We do have a lot more cases where left alignment fits better.

Thanks for the feedback! I changed it 👍

packages/component-library/src/components/form/mt-number-field/mt-number-field.vue
```diff
@@ -398,6 +408,12 @@ export default defineComponent({
     }
   }
 }
+
+.mt-number-field {
+  input.mt-number-field__align-end {
+    text-align: end;
+  }
+}
```
Haberkamp, line 416: I guess it makes sense to only use `.mt-number-field__align-end` as the selector as the extra scoping to `.mt-number-field` is unnecessary. It's already included in the longer selector.

Haberkamp (changes requested):

Haberkamp (approved):

seggewiss (approved):

PR #413, opened 2024-12-03
Add rule for borders

alastair-simon:
## What?
This rule checks every border value and reports violations.

## Why?
To catch incorrect or improper uses of border properties

## How?
The rule goes over each individual value and reports any violations.

## Testing?
I've written a few tests to check the rule works as expected


Haberkamp (changes requested):

Haberkamp (changes requested):
Also please add a changelog

Haberkamp (approved):

Haberkamp (approved):

PR #426, opened 2024-12-10
Rework tabs

Haberkamp:
## What?

This PR re-works the `mt-tabs` component.

## Why?

The current implementation had a bunch of a11y issues. And still relied on the options API.

## How?

I've re-written the tab component from scratch.

## Testing?

I've written a handful of unit tests to make sure the component works as expected.

## Anything Else?

Can only be merged after the popover component has been updated.


packages/component-library/src/components/overlay/mt-bare-popover/mt-bare-popover.vue
```diff
@@ -0,0 +1,177 @@
...
+      >
+
+      <ul class="mt-bare-popover__list">
+        <slot name="default" v-bind="{ closePopover: () => (isVisible = false) }" />
+      </ul>
+    </div>
+  </teleport>
+</template>
+
+<script setup lang="ts">
```
jleifeld, line 31: Just saw it randomly. Why do you create a new `mt-bare-popover` component instead of using `mt-popover`. The `mt-popover` has the exact same functionality like this component?

aragon999:
Is this planned for the 6.7 release?

I just noticed a problem with the `mt-tabs` component, namely if one is on a route of a tab item, not the correct tab is activated. This is done in the `sw-tabs` component by the following code: https://github.com/shopware/shopware/blob/751f8b579a6586df6ed1df006a803e6d851b2beb/src/Administration/Resources/app/administration/src/app/component/base/sw-tabs-item/index.js#L192-L222

Weltraumakustik:
What’s the state here?

PR #431, opened 2024-12-11
Update border token linting

alastair-simon:
## What?
Update the stylelint package for border properties to include calc

## Why?
Calc is used in a number of components css 

## How?
Added calc to global valid values 

## Testing?
I've added two additional tests


packages/stylelint-plugin-meteor/src/rules/prefer-border-token/prefer-border-token.test.ts
```diff
@@ -86,5 +90,14 @@ testRule({
...
     },
+    {
+      code: ".a { border-bottom-right-radius: 1px calc(2px / 0.18); }",
+      message:
+        'Unexpected hard-coded border radius of "1px" (meteor/prefer-border-token)',
+      line: 1,
+      column: 6,
+      endLine: 1,
+      endColumn: 55,
+    },
```
Haberkamp, line 101: Looks good, only think I would change is to rename `border-bottom-right-radius` to border-radius as the bottom right value can only accept one value.

Haberkamp (changes requested):

Haberkamp (approved):

PR #435, opened 2024-12-18
Improve colorpicker accessibility

alastair-simon:
## What?
Improve the accessibility of the mt-colorpicker component 

## Why?
Currently the colorpicker does not allow focus tabbing for navigation, making it in-accessable for users without a cursor.

## How?
Using the Focus-trap dependancy to allow focus navigation inside the colorpicker modal


Haberkamp (dismissed):
Please add test for your changes. It makes more sense to use unit testing instead of Storybooks play tests.

* Make sure it's possible to open an close the colorpicker via the keyboard
* Make sure it's possible to change values via the hue and alpha sliders
* Make sure it's possible to change the color via the big rectangular color field

For the big rectangular selection it makes sense to step in increments of one. And when pressing shift it can happen in 10 / 5 increments. Please also do this for the two sliders.

Lucsy3012 (dismissed):

alastair-simon:
Closing to resubmit with a cleaner branch structure

PR #441, opened 2025-01-03
fix: Change tooltip file to scss format

nguyenquocdaile:
## What?
If we run `pnpm run storybook` in component-library. It will have an error. 

Screenshot 2025-01-03 at 15 24 21

It requires a `sass` file, so I changed `tooltip.css` to a `tooltip.scss` file.

It seems that an SCSS file is required in the vite configuration here.

https://github.com/shopware/meteor/blob/main/packages/component-library/vite.config.ts#L31


jleifeld:
@Haberkamp can you have a look here. I remember you wanted to go away from SCSS. If this is the case we should fix the issue directly in the Vite config or the import

Haberkamp:
@nguyenquocdaile I do not experience the issue. The tooltip.css files does not contain any SCSS. Could you try pulling from main again and see if the issue still persists?

nguyenquocdaile:
> @nguyenquocdaile I do not experience the issue. The tooltip.css files does not contain any SCSS. Could you try pulling from main again and see if the issue still persists?

@Haberkamp looks weird, I don't know why it happened with me :D 
I am already in the main branch with the latest now 

https://github.com/user-attachments/assets/118171a3-ef3f-426c-ade9-d96f9aeec1f7

PR #443, opened 2025-01-03
Fix stylelint border rule

alastair-simon:
## What?
Adjustments to the stylelint border rule to allow expressions

## Why?
Currently the css property border-radius cannot accept expressions and throws a linting error

## How?
Adding expression to acceptable values for border radius

## Testing?
I have added some tests to make sure the adjustment is working


.changeset/dirty-nails-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-ag/stylelint-plugin-meteor": minor
```
Haberkamp, line 2: Bug fixes are always patch updates.

Haberkamp (changes requested):
There's one thing you can change the rest looks good.

.changeset/dirty-nails-heal.md
alastair-simon: Done

Lucsy3012 (approved):

Haberkamp (approved):

PR #444, opened 2025-01-06
fix: Emit focus event for mt-text-field component

nguyenquocdaile:
## What?
With current mt-text-field component. Emit a focus event doesn't work. It will led to that `sw-confirm-field` can't create a category

https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/app/component/form/sw-confirm-field/sw-confirm-field.html.twig#L27

image

## Solution 
Screenshot 2025-01-06 at 17 52 28


jleifeld (changes requested):
Looks good. Can you also write tests for this?

packages/component-library/src/components/form/mt-text-field/mt-text-field.interactive.stories.ts
```diff
@@ -230,3 +230,28 @@ export const VisualTestInheritanceActive: MtTextFieldStory = {
...
+    const canvas = within(canvasElement);
+    const input = canvas.getByPlaceholderText(args.placeholder);
+
+    // Listen for the focus event
+    const focusEvent = new Promise((resolve) => {
+      input.addEventListener("focus", resolve);
+    });
+
+    // Simulate focus event
+    await fireEvent.focus(input);
```
jleifeld, line 249: This doesn't test the real handleFocus behavior of the component. The test creates a event listener on it's own and triggers it directly. Then it checks afterwards if a Promise is defined (which is the case everytime).

Better is to test the real implementation:
1. User clicks on the input
2. Check if the component emits the "focus" event (you can check this directly in Storybook testing with args, see https://github.com/shopware/meteor/blob/main/packages/component-library/src/components/table-and-list/mt-data-table/mt-data-table.interactive.stories.ts#L174 )

Also it would be useful to change the name to `VisualTestHandleFocus` so that it generates a snapshot. Then the focus class is also tested

jleifeld (changes requested):

packages/component-library/src/components/form/mt-text-field/mt-text-field.interactive.stories.ts
```diff
@@ -230,3 +230,28 @@ export const VisualTestInheritanceActive: MtTextFieldStory = {
     isInherited: true,
   },
 };
+
+export const TestHandleFocus: MtTextFieldStory = {
```
Haberkamp, line 234: You could write the story like this:

```typescript
import { defineStory } from "@/_internal/story-helper";

export const TestHandleFocus = defineStory({
  name: "Handle focus",
  args: {
    placeholder: "Focus here",
  },
  async play() {
    await userEvent.tab();

    expect(args.MY_EVENT).toHaveBeenCalledOnce();
  }
})
```

jleifeld (approved):

PR #456, opened 2025-01-16
UX-59 - add accessibility state labels to switch component

alastair-simon:
## What?
Add aria attributes to mt-switch 

## Why?
Mt-switch was lacking aria attributes which improve the accessibility experience for users 

## How?
Added aria-checked, aria-label and role to input element inside mt-switch

### DIFF
--- a/packages/component-library/src/components/form/mt-switch/mt-switch.vue
+++ b/packages/component-library/src/components/form/mt-switch/mt-switch.vue
@@ -10,6 +10,9 @@
             :name="formFieldName || identification"
             :checked="inputState"
             :disabled="isDisabled"
+            :aria-checked="inputState"
+            :aria-label="label"
+            role="switch"
             @change.stop="onChange"
           />
           <div class="mt-field__switch-state">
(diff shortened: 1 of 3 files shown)


Lucsy3012 (approved):

packages/component-library/src/components/form/mt-switch/mt-switch.vue
```diff
@@ -10,6 +10,9 @@
             :name="formFieldName || identification"
             :checked="inputState"
             :disabled="isDisabled"
+            :aria-checked="inputState"
```
Haberkamp, line 13: Could you please write some tests to make sure your changes work? You can use the `toBeChecked` assertion of the [`jest-dom`](https://www.npmjs.com/package/@testing-library/jest-dom#tobechecked) package.

Haberkamp (changes requested):

Haberkamp (approved):

Lucsy3012 (approved):

PR #468, opened 2025-01-21
[Arvion] fix: upgrade package to new version

anthogez:
## What?

[Arvion] fix: upgrade package to new version.

It requires also "date-fns" to be updated, otherwise there will be resolution conflict between "date-fns" and "date-fns-tz"

Solving breaking changes that dependabot failed here
https://github.com/shopware/meteor/pull/394


Haberkamp:
@anthogez Thanks for the PR, could you update the pnpm-lock file too, otherwise the build won't pass

anthogez:
> @anthogez Thanks for the PR, could you update the pnpm-lock file too, otherwise the build won't pass

@Haberkamp Apologies, I'll have this done. The engine that I am building to automate this is still missing the relock 😅 
I guess I'll have to push this manually. Thanks!

Haberkamp (approved):

PR #481, opened 2025-01-23
feat(stylelint-plugin-meteor): auto-fix rem spacing values

Haberkamp:
## What?

I added a auto-fixer for rem spacing values.

## Why?

This makes it easier to migrate a large amount of spacing over to the spacing tokens.

## How?

The fixer multiplies the rem value by 16, checks if the value is on the scale. If it is then it replaces that with a spacing token. If not it just reports that line.

## Testing?

I've added some tests to make sure the rule works as expected.


packages/component-library/src/components/layout/mt-card/mt-card.vue
```diff
@@ -177,7 +177,7 @@ const cardClasses = computed(() => ({
 
 .mt-card {
   max-width: 60rem;
```
Lucsy3012, line 179: @Haberkamp Should the 60rem stay without a variable?

Lucsy3012 (approved):

alastair-simon (approved):

PR #483, opened 2025-01-24
fix(component-library): fix ts node_module resolution

ffrank913:
## What?
Added node_module paths for correct TS resolution for component-library.

## Why?
Desired packages in component-library don't have the correct resolution to root's node_modules-folder.

## How?
By adding both relevant node_modules-folders to the component-library's ts-config.

## Testing?
This change is only relevant for developing convenience so no testing is needed.


Haberkamp:
@ffrank913 thanks for the PR, can you explain what this change does in more detail? I never experience any issues with resolving node modules.

PR #492, opened 2025-02-03
feat: add `contextCompareShopwareVersion` to SDK

tajespasarela:
## What?
Added a new feature `contextShopwareVersion` to meteor SDK

## Why?
To enhance the version comparison functionality by supporting versions that have four numeric parts and suffixes.

## How?
- Added `contextShopwareVersion` sender to the context SDK.

### DIFF
--- a/.changeset/fluffy-teachers-complain.md
+++ b/.changeset/fluffy-teachers-complain.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-admin-sdk": minor
+---
+
+Add `contextCompareShopwareVersion` to sdk to compare the current Shopware version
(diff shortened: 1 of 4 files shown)


jleifeld (changes requested):
Add documentation for this, otherwise looks good

jleifeld (approved):

seggewiss (approved):

SebastianFranze:
Hello @tajespasarela and sorry for being late to the party.

This implementation is not going to work. `context.compareShopwareVersion` must not depend on the version of Shopware. Implementing it with a `postMessage` approach adds this dependency because the message is only handled by shops with a specific version.

Hence, to use `context.compareShopwareVersion` you have to check the Shopware version before using it :shrug: 

A pseudo Example:

```js
async function getVersionCompare() {
     const shopwareVersion = await context.getShopwareVersion();

    if (shopwareVersion <= 6.7.0.0)
       return context.compareShopwareVersion;

    return customVersionHandling;    
}
```

PR #507, opened 2025-02-17
chore: Export type Toast so that it can be used when working with MtToast components

SebastianFranze:
## What?

With this PR we make the `interface Toast` from `components/feedback-indicator/mt-toast/mt-toast.vue` available outside of the component library.

## Why?

Right now, the `interface`  cannot be used when working with the `MtToast` component.

### DIFF
--- a/.changeset/thick-ligers-mix.md
+++ b/.changeset/thick-ligers-mix.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": minor
+---
+
+chore: export type Toast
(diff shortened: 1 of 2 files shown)


Haberkamp (changes requested):
Looks great, one small thing. Please add a changelog file with the type: minor using `npx changeset` and describe your change.

SebastianFranze:
@Haberkamp 

I hope the changelog is fitting. It's my first time contributing to meteor :sweat_smile:

Haberkamp (approved):

PR #513, opened 2025-02-20
fix: remove duplicated remove-inheritance event

jleifeld:
## What?

The `remove-inheritance` event gets triggered twice on click in `mt-switch` which leads to issues

## Why?

Because the original `remove-inheritance` gets already emitted with `$attrs`

## How?

Remove the other one

## Testing?

Tested it carefully in the Shopware platform

### DIFF
--- a/packages/component-library/src/components/form/mt-switch/mt-switch.vue
+++ b/packages/component-library/src/components/form/mt-switch/mt-switch.vue
@@ -30,7 +30,6 @@
           :is-inheritance-field="isInheritanceField"
           :is-inherited="isInherited"
           @inheritance-restore="onInheritanceRestore($event)"
-          @inheritance-remove="$emit('inheritance-remove', $event)"
         >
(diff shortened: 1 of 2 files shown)


packages/component-library/src/components/form/mt-switch/mt-switch.vue
```diff
@@ -30,7 +30,6 @@
           :is-inheritance-field="isInheritanceField"
           :is-inherited="isInherited"
           @inheritance-restore="onInheritanceRestore($event)"
```
seggewiss, line 32: What about the restore event?

jleifeld: Restoring the inheritance didn't lead to issues. I just want to avoid changes here which can break other stuff. So these are only the necessary fixes to make it work.

But after the major we should look at all foundings and try to improve it

PR #514, opened 2025-02-20
Add tests for mt switch

Haberkamp:
## What?

This PR adds a handful of tests for the `mt-switch` component.

## Why?

Adding more tests increases our confidence that things work and we reduce the number of bugs and regressions.

## How?

I added a handful of tests and then did some refactoring.

## Testing?

I also added a lot of unit tests to make sure the component works the same way it did before

## Anything Else?

* I added a v-model for the `mt-switch` component
* I made the switch field requirable
* I migrated the component over to the composition api
* The component now uses plain CSS
* I removed the coupling from the `mt-text-field`


packages/component-library/src/components/form/mt-switch/mt-switch.vue
```diff
@@ -1,394 +1,206 @@
...
+      ]"
+    >
+      <input
+        :checked="isInherited ? inheritedValue : checked || modelValue"
+        type="checkbox"
+        :class="['mt-switch__switch', { 'mt-switch__switch--errored': !!error }]"
+        :id="id"
+        :disabled="disabled || isInherited"
+        :required="required"
+        :aria-describedby="!!error ? errorId : undefined"
```
jleifeld, line 28: Can you add also a `aria-label` here? Otherwise many selectors in tests from my side would break and a hard to replace because I need to find the input somewhere close to the correct label

jleifeld (dismissed):

alastair-simon (approved):

PR #526, opened 2025-02-21
catch failed dynamic imports in mt-icon

bschulzebaek:
## What?
Failure `mt-icon` dynamic import is not caught, which leads to runtime exceptions if using non-existent icons.

## Why?
A failed import shouldn't break the app, especially thinking of adding/removing icons, using arbirtrary names in tests, etc.


Haberkamp (changes requested):
Thanks for the PR! Could you create a [changelog](https://github.com/shopware/meteor/blob/main/CONTRIBUTING.md#create-a-changelog)? And this PR needs a test to make sure it works and I won't break it in the future. 😅 

When rendering the component please use `@testing-library/vue` and not `@vue/test-utils`. Testing library is designed to avoid testing implementation details as much as possible.

bschulzebaek:
@Haberkamp sure, I'll pick this up once we're done with the meteor migration ✌🏻

Haberkamp (approved):

alastair-simon (approved):

PR #527, opened 2025-02-24
fix: update mt-select popover position

quando1910:
## What?
- mt-select popover position is incorrect (always on the top) when the select box is in the bottom view.
- the popover is also incorrect when scroll the view.
## Why?
- The position of popover should be `fixed` instead of `absoluted`
- Besides, the `vnodeContext?._uid` in function registerVirtualScrollingElement is always return undefined, it should be `vnodeContext?._?.uid`

## Screenshots (optional)
image
image


Haberkamp (changes requested):
Please add a changelog using `npx` changesets. For a full guide you can read the `CONTRIBUTING.md` file

quando1910:
> Please add a changelog using `npx` changesets. For a full guide you can read the `CONTRIBUTING.md` file

Hi, I already added

Haberkamp (approved):

alastair-simon (approved):

PR #533, opened 2025-02-25
feat(component-library): make mt-url-field component public

bschulzebaek:
## What?

Made the `mt-url-field` component public, which is not exported by default currently.


seggewiss (approved):

Haberkamp (changes requested):
Please add a changelog. That way the package version automatically jumps to a new minor version.

Haberkamp (approved):

PR #543, opened 2025-02-26
Simplify icon component

Haberkamp:
## What?

I added simpler way to use icons.

```html
<!-- Old way -->
<mt-icon name="regular-3d" />

<!-- New way: Regular icon -->
<mt-icon name="3d" />

<!-- New way: Solid icon -->
<mt-icon name="3d" mode="solid" />
```

## Why?

In 90% percent of the case we use regular icons instead of the solid ones. 

I also got the feedback multiple times, that it's weird or annoying always needing to define the mode i.e. regular or solid.

## How?

I designed the component in a way where the old and the new way both work. Over time we're going to deprecate the new way, and also provide an easy automated way to migrate to the new version.

## Testing?

I wrote a couple of test to make sure the old and the new way work.


Lucsy3012 (approved):

packages/component-library/src/components/icons-media/mt-icon/mt-icon.vue
```diff
@@ -57,8 +59,10 @@ const styles = computed(() => {
 watch(
   () => props.name,
   (newName) => {
-    const [variant] = newName.split("-");
-    const iconName = newName.split("-").slice(1).join("-");
+    const variant = /^(solid|regular)/.test(newName) ? newName.split("-")[0] : props.mode;
```
bschulzebaek, line 62: Should we be using regex in this component? I know it's a quite simple one, but imagine having dozens of icons being rendered at once times 3 RegEx per icon, could be interesting performance-wise.

bschulzebaek: Or just use `split('-.)`: If you have two items, it's safe to assume that both `variant` and `name` are defined

Haberkamp: Ok, I added a computed property and replaced the regexes with `.startsWith()`

bschulzebaek (approved):

Lucsy3012 (approved):

PR #547, opened 2025-02-26
feat (NEXT-40858): Add message type windowGetId

SebastianFranze:
## What?

Add new message type windowGetId

## Why?

We want to retrieve a unique identifier for Amin windows


packages/admin-sdk/src/window/index.ts
```diff
@@ -36,3 +37,7 @@ export type windowRouterPush = {
 export type windowReload = {
   responseType: void,
 }
+
+export type windowGetId = {
```
jleifeld, line 41: Add a comment for explanation what this does

jleifeld (changes requested):
Can you add this method also in the documentation with this PR?

docs/admin-sdk/docs/guide/2_api-reference/window.md
```diff
@@ -65,3 +65,34 @@ No parameters required.
 
 #### Return value:
 Returns a promise without data.
+
+### Get an unique identifier for the window
+
+> Available since Shopware v6.7.1.0
```
SebastianFranze, line 71: Is 6.7.0.0 already feature complete? :thinking:

jleifeld: Almost. I would say we merge the documentation with v.6.7.1.0 information. If we can get this feature merged before than it is easy to change the version back to 6.7. Other way around is difficult

jleifeld (approved):

Haberkamp (approved):

PR #564, opened 2025-03-04
feat: add compare Shopware Version logic directly in the SDK

tajespasarela:
## What?
This PR adds the logic for compare Shopware version directly in SDK

## Why?
To make it compatible with any version of Shopware

## How?
the comparison is done directly in the SDK

## Testing?
Added test for the comparison logic

### DIFF
--- a/.changeset/little-moose-accept.md
+++ b/.changeset/little-moose-accept.md
@@ -1,0 +1,6 @@
+---
+"@shopware-ag/meteor-admin-sdk": minor
+---
+
+- Changed name of `context.compareShopwareVersion` to `context.compareIsShopwareVersion` 
+- Changed order of parameters of `context.compareIsShopwareVersion`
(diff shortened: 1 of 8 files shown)


jleifeld (dismissed):

SebastianFranze:
Maybe we should not reinvent the wheel. We can use the `semver` package to compare the versions. Our only problem is that we have a four digit version with a leading `6.`. After checking if this is present, we can pass it to semver:

```ts
import { valid, gte. lte, [...] } from 'semver';

type Comparator = '=' | '>' | '<' | '<=' | '>=';

export default function getCompareShopwareVersion(
  getShopwareVersion: () => Promise<string>
) {
return async (versionToCompare: string, comparator: Comparator): Promise<boolean> => {
  const shopwareVersion = await getShopwareVersion();

  if (!valid(shopwareVersion) && shopwareVersion.match(/^6\.[\d]+.[\d]+.[\d]+/)) {
    shopwareVersion = shopwareVersion.substring(2);
  }

  switch (comparator) {
    case '<=':
      return gte(shopwareVersion, versionToCompare);
    case '=':
      return ...
    default:
      throw new Error('invalid comparator');
  }
}
```
If it is still invalid we throw an error.

seggewiss (approved):

PR #575, opened 2025-03-06
Wrap text in tooltip

Haberkamp:
## What?

This PR wraps the content of tooltips and defines a max-width for help texts.

## Why?

A lot of times the content of tooltips gets very long. To prevent them from overflowing we a) wrap the content
and b) define a max width for help texts.

## How?

The tooltip now wraps the content.

## Testing?

I added visual test to make sure the content gets wrapped.

## Anything Else?

It's now possible to define the max width of a tooltip using the `max-width` prop.


packages/component-library/src/components/overlay/mt-tooltip/mt-tooltip.vue
```diff
@@ -202,6 +203,7 @@ provide(TooltipContext, true);
   padding: var(--scale-size-12);
   border-radius: var(--border-radius-overlay);
   width: max-content;
+  word-break: break-all;
```
fabianhueske, line 206: `word-break: break-all;` could cause a lot of problems, since it's breaking everything everywhere. Could you try using something like `overflow-wrap`? https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap Correct me if stupid 😄

Haberkamp: Yes, that works better. I'll change that.

PR #578, opened 2025-03-06
fix: mt-single-select truncation issue

jleifeld:
## What?

The mt-select in single selection mode gets truncated too fast

## How?

Don't use the pills, just use the input value


packages/component-library/src/components/form/_internal/mt-select-base/_internal/mt-select-selection-list.vue
```diff
@@ -167,6 +168,18 @@ export default defineComponent({
...
       return this.alwaysShowPlaceholder || this.selections.length === 0 ? this.placeholder : "";
     },
+
+    inputValue(): string {
+      if (this.multiSelection) {
+        return this.searchTerm;
+      }
+
+      console.log("this.selections", this.selections);
+      console.log("this.labelProperty", this.labelProperty);
```
Haberkamp, line 178: Those need to be removed

Haberkamp (changes requested):

jleifeld:
We have still problems with the `small` prop to true. I will fix this also

jleifeld:
> We have still problems with the `small` prop to true. I will fix this also

Is fixed. Now I will test the changes also in platform to make sure nothing breaks

Haberkamp (approved):

PR #593, opened 2025-03-11
feat: Remove duplicated external link icon; Make mt-link more responsive

SebastianFranze:
## What?

* Added `mt-link--external` and `mt-link--internal` CSS classes to `mt-link` component
* Excluded `mt-link--external` and `mt-link--internal` in `global.scss` to exclude them from adding an external link icon
* Removed static font sizes for `mt-link`
* Changed icon size and gap to `em` values to provide more responsiveness.

## Why?

When passing  `target="_blank"` to `mt-link`, the global CSS adds an `after` element to display an external link icon. The `mt-link` component already displays such an icon. Hence, the icon was displayed twice.

Additionally, I removed the fixed font-size for the `mt-link` component and replaced `rem` scales with `em` scales.

## Screenshots (optional)

Before:

image

After:

image


packages/component-library/src/components/assets/scss/global.scss
```diff
@@ -38,11 +38,16 @@ body {
 }
 
 /* style inline external links */
-a[target="_blank"]:not(.mt-internal-link, .mt-button) {
+a[target="_blank"]:not(
```
bojanrajh, line 41: Not related to this PR, but this is one of the most annoying selectors in the whole bundle. Can we somehow limit it to a shopware backend, like `.sw-desktop a...` or `.sw-page a...`? @Haberkamp

Haberkamp (commented):
@SebastianFranze have you tested the changes in the admin to see if this breaks something? In theory it should not, but you never know for sure.

packages/component-library/src/components/navigation/mt-link/mt-link.vue
```diff
@@ -6,6 +6,7 @@
       `mt-link--${variant}`,
       {
         'mt-link--disabled': disabled,
+        [`mt-link--${type}`]: type !== undefined,
       },
     ]"
     :href="disabled ? undefined : to"
```
SebastianFranze, line 12: The link component does not add a `target=_blank` to external links. Should we add it?

SebastianFranze:
@Haberkamp I checked and in the Admin we do not use the `mt-link` component at all. The only place It is used is in the `sw-gtc-checkbox` component but it seems like the component is not globally available. The component is not resolved as you can see in the screenshot :shrug: 

image

Additionally, we have the styles "duplicated" in the Admin's `global.scss`,  which will probably override my changes in the Admin.

https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/app/assets/scss/global.scss#L61

Haberkamp:
> @Haberkamp I checked and in the Admin we do not use the `mt-link` component at all. The only place It is used is in the `sw-gtc-checkbox` component but it seems like the component is not globally available. The component is not resolved as you can see in the screenshot 🤷
> 
> image
> 
> Additionally, we have the styles "duplicated" in the Admin's `global.scss`, which will probably override my changes in the Admin.
> 
> https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/app/assets/scss/global.scss#L61

@SebastianFranze You could include the `.mt-external-link` class in that `:not()` selector. So, the styles no longer got overwritten.

packages/component-library/src/components/navigation/mt-link/mt-link.vue
Haberkamp: I would leave it like it is right now. With the default browser behaviour we allow the user to choose if they want to open the link in the same tab or to press CTRL to open it in a new tab. 

If we use `taget="_blank"` we force the user to open the tab in a new page.

[More reading](https://css-tricks.com/use-target_blank/)

SebastianFranze:
Because of the global styles in the Admin nothing should change. We should clean it up after we released this changes. See https://github.com/shopware/shopware/issues/7887

tajespasarela (approved):

jleifeld (approved):

PR #598, opened 2025-03-17
fix: slider input fields visual improvement

Deristes:
### `mt-slider` component issues:
This pull request includes several changes to the `mt-slider` component in the `component-library` package to improve its styling and functionality. The most important changes include adding a computed property for dynamic input field width and updating styles for better alignment.

## Why?
The current input field is limited in size, accommodating only two digits, which can result in numbers being cut off. This limitation affects usability, especially when users need to input larger values. By enhancing the input field's width dynamically, we can ensure that all numbers are displayed correctly and improve the overall user experience.

### Enhancements to `mt-slider` component:
* Added a computed property `inputFieldWidth` to dynamically calculate the width of the input fields based on the length of the `min` and `max` values.
* Applied the computed `inputFieldWidth` to both left and right number fields to ensure consistent sizing.

## Screenshots (optional)
image
image

## Anything Else?
This is linked to: shopware/shopware#7560 and [NEXT-41037](https://shopware.atlassian.net/browse/NEXT-41037)
Please provide critical feedback to this PR about the solution!

[NEXT-41037]: https://shopware.atlassian.net/browse/NEXT-41037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/component-library/src/components/form/mt-slider/mt-slider.vue
```diff
@@ -381,6 +383,11 @@ export default defineComponent({
             transform: "translateX(50%)",
           };
     },
+
+    inputFieldWidth(): string {
+      const chars = Math.max(this.max.toString().length, this.min.toString().length) + 2;
```
dfrancos-hub, line 388: Hi @Deristes 👋 , would this set always both min (left input) and max (right input) to the same size?. If so maybe we could make each one adjusted to their own max size, so in the cases where min value is limited by a larger minDistance, the min left input field witdh is adjusted separately (and viceversa in case of more chars in the min value with negative).

packages/component-library/src/components/form/mt-slider/mt-slider.vue
```diff
@@ -33,6 +33,7 @@
         size="small"
         :number-type="step % 1 === 0 ? 'int' : 'float'"
```
dfrancos-hub, line 34: Regarding decimals, what happens if the max input chars are 3 (for a max of 100), although the step allows something like 99.99?

dfrancos-hub (changes requested):

packages/component-library/src/components/form/mt-slider/mt-slider.vue
Deristes: Good point, will add that :+1:

packages/component-library/src/components/form/mt-slider/mt-slider.vue
Deristes: The left and right input should be the same size. The Inputs are *not* there to set min and max. 
For example the range is 0 ... 2000 and than i select 1000 - 2000 i also need 4 chars in the left input field even tho min only needs one.
Does this answer this thread?

dfrancos-hub: mmm, you are right, although I was refering to the case of having a range between 0 and 220, but having a minDistance of 150, that would mean the real max value for the min input is 70, no?

dfrancos-hub: just ℹ️ we had a call and concluded to keep both the same size in order to keep the symmetry, regardless of the calculated max chars of each input 😄

dfrancos-hub (approved):

PR #599, opened 2025-03-18
fix: build process, @use instead of @import for scss

BrocksiNet:
## What?
- Fix the build command in this repository
- Do not use the @import from SCSS anymore because it is deprecated

## Why?
- We should use up-to-date SCSS functions and not the old stuff

## Testing?
- Yes, this should be tested by someone who is working with Meteor more regular

## Anything Else?
- There are still a lot of SCSS warnings and errors that could be fixed
  - For example lighten and darken functions are also deprecated


Haberkamp:
@BrocksiNet Thanks for the PR. I'm closing is because we've decided to move off from SCSS.

PR #600, opened 2025-03-18
fix: password flickering on submit

BrocksiNet:
## What?
- Key ENTER will throw a submit event on the input password
- Added `click.prevent` to not toggle button on bubble

## Why?
https://github.com/shopware/shopware/issues/7523

## Testing?
- We need this PR merged
- And [the one](https://github.com/shopware/shopware/pull/7664) in the Platform to make everything work

## Anything Else?
- I think it is a security issue to have all the JS (from every plugin/app) on the login page

### DIFF
--- a/packages/component-library/src/components/form/mt-password-field/mt-password-field.vue
+++ b/packages/component-library/src/components/form/mt-password-field/mt-password-field.vue
@@ -42,7 +42,7 @@
 
     <mt-field-error v-if="!!error" :error="error" />
 
-    <div class="mt-password-field__hint">
+    <div v-if="$slots.hint" class="mt-password-field__hint">
       <slot name="hint" />
     </div>
   </div>
@@ -81,9 +81,9 @@ defineEmits<{
 }>();
 
 defineSlots<{
-  prefix?;
-  suffix?;
-  hint?;
+  prefix?: unknown;
+  suffix?: unknown;
+  hint?: unknown;
 }>();
 
 const id = useId();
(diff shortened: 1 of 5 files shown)


packages/component-library/src/components/form/mt-password-field/mt-password-field.vue
```diff
@@ -22,7 +23,8 @@
 
       <button
         v-if="toggable"
-        @click="showPassword = !showPassword"
+        type="button"
```
seggewiss, line 26: Why is type button needed here? It should not be necessary nor will this change the semantics for Screen Readers as it is already a button 🤔

seggewiss (changes requested):

Haberkamp (changes requested):
Please add a unit test, just to make sure we don't create a regression.

packages/component-library/src/components/form/mt-password-field/mt-password-field.vue
BrocksiNet: If your buttons are not for submitting form data to a server, be sure to set their type attribute to button . Otherwise, they will try to submit form data and to load the (nonexistent) response, possibly destroying the current state of the document.

packages/component-library/src/components/form/mt-password-field/mt-password-field.vue
```diff
@@ -40,7 +42,7 @@
 
     <mt-field-error v-if="!!error" :error="error" />
 
-    <div class="mt-password-field__hint">
+    <div v-if="$slots.hint" class="mt-password-field__hint">
```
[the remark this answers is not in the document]
BrocksiNet: Just to have a clean DOM. We can also remove it.

BrocksiNet:
@Haberkamp, the `pnpm run test:unit` is not running any tests in my local.
Could you be so kind and add the test for me? Thx.

keulinho (approved):

seggewiss (approved):

Haberkamp (approved):

Issue #594, opened 2025-03-12
Add monthly and yearly subscription options in one tier


Bird87ZA:
👥 USER STORY

**As a** extension partners
**I want** different subscription models
**so that** merchants can have various pricing models to choose from

ℹ BACKGROUND INFORMATION

Currently the implementation for IAP's only allow one pricing model to be available to the merchant. We need to provide some options here so both merchants and extension partners have more variety

This is the SBP ticket: https://shopware.atlassian.net/browse/ECTO-1817

✅ ACCEPTANCE CRITERIA

WHEN a In App Purchase is initiated THEN the trigger for iapCheckout should accept various pricing models

PR #601, opened 2025-03-18
feat: added new pricemodels for iaps

Bird87ZA:
## What?
Added pricing models to IAP's

## Why?
Expands the list of options merchants can choose from and partners can use

## How?
Added some extra fields

## Testing?
Will be tested via platform

## Anything Else?
fixes https://github.com/shopware/meteor/issues/594

### DIFF
--- a/packages/admin-sdk/src/iap/index.ts
+++ b/packages/admin-sdk/src/iap/index.ts
@@ -9,4 +9,5 @@ export type iapCheckout = {
     responseType: unknown,
 
     identifier: string,
+    variant?: 'monthly' | 'yearly' | 'consumable' | 'non-consumable',
 }


packages/admin-sdk/src/iap/index.ts
```diff
@@ -5,8 +5,9 @@ import { createSender } from '../channel';
  */
 export const purchase = createSender('iapCheckout');
 
-export type iapCheckout = {
+export type iapCheckout<T extends 'monthly' | 'yearly' | 'consumable' | 'non-consumable' = 'monthly' | 'yearly' | 'consumable' | 'non-consumable'> = {
     responseType: unknown,
 
     identifier: string,
+    variant: T,
```
cyl3x, line 12: Since no second property depends on another, you can inline the type. Also, for backwards compatibility, the property shouldn't be required:
```suggestion
export type iapCheckout = {
    responseType: unknown,

    identifier: string,
    variant?: 'monthly' | 'yearly' | 'consumable' | 'non-consumable',
```

cyl3x (approved):

En0Ma1259 (approved):

lernhart (approved):

mstegmeyer (approved):

Bird87ZA:
This is not required anymore. See https://github.com/shopware/SwagExtensionStore/pull/55.

Short explanation: The modal handles the variant selection. Meaning, when a merchant wants to subscribe to a feature, the modal will open up with the price models for this feature.

The code here only triggers the opening of the modal, not the type of price model that should be charged.

PR #605, opened 2025-03-19
Reduce flakyness of checkbox tests

Haberkamp:
## What?

This PR reduces the flakiness of an interaction test.

## Why?

We had a flaky test for the `mt-checkbox` component. It made sure that a tooltip is visible when the user hovers over the inheritance toggle of the tooltip. 

## How?

I removed the assertion from the interaction test, but added a couple of unit tests to make sure it still works.

## Testing?

When the unit tests pass then everything works as expected.

## Anything Else?

* I added the missing `isInherited` property to the `mt-checkbox` component.


packages/component-library/__snapshots__/interaction-tests-form-mt-checkbox--visual-test-inherited-snap.png
Haberkamp, line 1: It makes sense that the tooltip is no longer visible. We hide the tooltip when we either click on the button or press enter or space on the button.

Lucsy3012 (approved):

alastair-simon (approved):

Issue #615, opened 2025-03-21
[Bug]: Popovers not visible in modals


marcelbrode:
### Package

@meteor/component-library

### Package version

4.8.0

### Link to minimal reproduction

-

### Steps to reproduce

- Go anywhere in the admin where helpText is in a modal.
  - For example: CMS -> image slider block -> config
- Hover over helpText icon

### Actual behavior

- You see nothing.
- Sometimes, you see a helpText behind the modal, if would surpass its borders

![Image](https://github.com/user-attachments/assets/efd2dc57-507b-4ba7-b0be-ca9a309806c8)

### Expected behavior

See popovers above everything else

### System Info

```shell

```

### Code of Conduct

- [x] I have read and agree to follow Meteor's [Code of Conduct](https://github.com/shopware/meteor/blob/main/CODE_OF_CONDUCT.md)

PR #617, opened 2025-03-24
fix: Tooltip in modals

seggewiss:
## What?
Fixes https://github.com/shopware/meteor/issues/615

## Screenshot
image


alastair-simon (dismissed):

packages/component-library/src/components/form/_internal/mt-base-field/mt-base-field.vue
```diff
@@ -261,6 +261,7 @@ $mt-field-transition:
   .mt-field__help-text {
     margin-left: var(--scale-size-4);
     align-self: center;
+    z-index: 2100;
```
Haberkamp, line 264: Shouldn't the inline style in `mt-tooltip` be enough? Or am I missing something here?

And please add a changelog to this PR. The rest looks good.

Haberkamp (changes requested):

packages/component-library/src/components/form/_internal/mt-base-field/mt-base-field.vue
seggewiss: Done 👍

Haberkamp (approved):

alastair-simon (approved):

PR #630, opened 2025-03-26
Migrate styles of mt-select-result-list to plain css

Haberkamp:
## What?

This PR migrates the `mt-select-result-list` over to plain css.

## Why?

SCSS does not provide much, if even at all, value for us.

## How?

I migrated the SCSS to plain CSS.

## Testing?

If the visual tests pass everything is fine.


alastair-simon (approved):

packages/component-library/src/components/form/_internal/mt-select-base/_internal/mt-select-result-list.vue
```diff
@@ -364,22 +360,22 @@ $mt-select-result-list-transition: all ease-in-out 0.2s;
   overflow-x: hidden;
   overflow-y: auto;
   border: 1px solid var(--color-border-primary-default);
-  box-shadow: 0 3px 6px 0 $color-gray-300;
+  box-shadow: 0 3px 6px 0 rgba(142, 142, 147, 0.16);
```
fabianhueske, line 363: We could add the new shadow token here.

PR #639, opened 2025-03-31
fix: mt-modal z-index

### LINKED ISSUE #635: [Bug]: mt-modal has too high z-index
### Package

@meteor/component-library

### Package version

4.8.0

https://github.com/shopware/shopware

### Steps to reproduce

Use current trunk with CMS-Extensions plugin and try to edir the settings of a product slider element in the CMS.

CMS Extensions overrides the `sw-cms-slot` and uses mt-modal.
The error wont occur without CMS Extensions, since the core still uses sw-modal.

Image
 

### Actual behavior

When clicking into the selects in the configuration, it opens behind the modal, because of the z-index. 
The mt-modal has z-index: 2000 and sw-modal has z-index: 1000

### Expected behavior

Select opens normally

### System Info

```shell

```

### Any additional comments?

_No response_

### Code of Conduct

- [x] I have read and agree to follow Meteor's [Code of Conduct](https://github.com/shopware/meteor/blob/main/CODE_OF_CONDUCT.md)

taltholtmann:
fixes #635 https://github.com/shopware/meteor/issues/635

## What?
mt-modal has too high z-index

## How?
Changed z-index. It has now the same as the old sw-modal

## Testing?
Testet in current trunk with cms extensions plugin installed - that plugin uses the mt-modal.


seggewiss:
Should be no longer necessary after https://github.com/shopware/meteor/pull/617

taltholtmann:
> Should be no longer necessary after #617. Or is there a general problem with z-indexes in meteor? 🤔

Yes, there is still a problem. The mt-modal has a z-index of 2000 and the popovers for the select lists have 1100. So they are hidden behind the modal.

Haberkamp (changes requested):
Please add a changelog using npx changesets. Then we're good to go and can merge it.

seggewiss:
@taltholtmann so you should also change the [tooltip z-index](https://github.com/shopware/meteor/blob/main/packages/component-library/src/components/overlay/mt-tooltip/mt-tooltip.vue#L26) also. 

@Haberkamp I would expect meteor to have a z-index strategy that we shouldn't brake to often as it's a pain for people using the comp library and could break their custom styling. Does something like this exist in the form of tokesn?

taltholtmann:
> @taltholtmann so you should also change the [tooltip z-index](https://github.com/shopware/meteor/blob/main/packages/component-library/src/components/overlay/mt-tooltip/mt-tooltip.vue#L26) also.

But the mt-tooltip has 1100. Which is okay, I guess

Haberkamp (approved):

seggewiss (approved):

PR #645, opened 2025-04-01
fix(component-library): mt-email-field showing error message on init

tajespasarela:
## What?
This PR fixes the `mt-email-field` that shows error message right away

## Why?
When user enters in an empty form, it shows errors in the email field.

## How?
It removes the validity check from the onMount of the component

## Testing?
Added test, and modify the failing tests because of the change

## Screenshots (optional)
image


Haberkamp (changes requested):
Looking good. One small thing. Could you add a changeset?

seggewiss (approved):

tajespasarela:
> Looking good. One small thing. Could you add a changeset?

done @Haberkamp 👍🏻

PR #654, opened 2025-04-03
Migrate select selection list to plain css

Haberkamp:
## What?

This PR migrates the `mt-select-selection-list-to-plain-css`

## Why?

For us, SCSS does not provide much value.

## How?

I replaced the SCSS with plain CSS.

## Testing?

You can do some manual testing if you want. Otherwise, if the visual tests pass. Then everything is fine.


packages/component-library/src/components/form/_internal/mt-select-base/_internal/mt-select-selection-list.vue
```diff
@@ -283,100 +283,94 @@ export default defineComponent({
...
+  margin: var(--scale-size-8) var(--scale-size-6) 0 0;
+}
+
+.mt-select-selection-list button.mt-select-selection-list__load-more-button {
+  padding: var(--scale-size-8) var(--scale-size-12);
+  margin: 0 var(--scale-size-6) 0 0;
+  color: var(--color-text-brand-default);
+  font-size: 12px;
+  line-height: 14px;
+  border-radius: 2px;
```
fabianhueske, line 324: Could we use variables here?

packages/component-library/src/components/form/_internal/mt-select-base/_internal/mt-select-selection-list.vue
```diff
@@ -283,100 +283,94 @@ export default defineComponent({
...
-    display: inline-block;
-    min-height: 46px;
-    padding: var(--scale-size-12) var(--scale-size-16) var(--scale-size-12) var(--scale-size-8);
-
-    &::placeholder {
-      color: lighten($color-darkgray-200, 25%);
-      white-space: break-spaces;
-    }
-  }
+  min-height: 32px;
```
fabianhueske, line 336:
```suggestion
  min-height: var(--scale-size-32);
```

packages/component-library/src/components/form/_internal/mt-select-base/_internal/mt-select-selection-list.vue
```diff
@@ -283,100 +283,94 @@ export default defineComponent({
...
+  padding: var(--scale-size-4) var(--scale-size-16) var(--scale-size-4) var(--scale-size-8);
 }
 
-.mt-field--medium .mt-select-selection-list {
-  .mt-select-selection-list__item-holder .mt-label {
-    margin: var(--scale-size-4) var(--scale-size-6) 0 0;
-  }
+.mt-select-selection-list .mt-select-selection-list__input {
+  display: inline-block;
+  min-height: 46px;
```
fabianhueske, line 342:
```suggestion
  min-height: calc(var(--scale-size-48) - var(--scale-size-2));
```

packages/component-library/src/components/form/_internal/mt-select-base/_internal/mt-select-selection-list.vue
```diff
@@ -333,13 +337,13 @@ export default defineComponent({
...
   .mt-select-selection-list__input {
-  min-height: 32px;
+  min-height: var(--scale-size-32);
   padding: var(--scale-size-4) var(--scale-size-16) var(--scale-size-4) var(--scale-size-8);
 }
 
 .mt-select-selection-list .mt-select-selection-list__input {
   display: inline-block;
-  min-height: 46px;
+  min-height: calc(var(--scale-size-48) + var(--scale-size-2));
```
fabianhueske, line 346: Should the calc operation be plus or minus? 🤔

packages/component-library/src/components/form/_internal/mt-select-base/_internal/mt-select-selection-list.vue
```diff
@@ -283,100 +282,89 @@ export default defineComponent({
...
+  color: var(--color-text-brand-default);
+  font-size: var(--font-size-xs);
+  padding: var(--scale-size-8) var(--scale-size-12);
+  height: var(--scale-size-32);
+  line-height: 1.1;
+  border: 1px solid var(--color-border-primary-default);
+  border-radius: var(--border-radius-xs);
+  background: var(--color-background-primary-default);
+
+  &:hover {
```
alastair-simon, line 327: There's still a scss selector here

fabianhueske: This would actually work since it's plain CSS nesting (same syntax in that case as SCSS). So it would work if we allow to use CSS nesting. Supported by most modern browsers.

.changeset/clean-rabbits-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": patch
+---
+
+Update styling of load more button in select
```
fabianhueske, line 5: Seems like there are two changes on this PR? Updating the load more button and moving to plain CSS? Should test the changes on the button 👀

packages/component-library/src/components/form/_internal/mt-select-base/_internal/mt-select-selection-list.vue
alastair-simon: You only have to go back a couple of versions of browsers and it isn't supported, I guess it's a trade off.

PR #657, opened 2025-04-03
feat: allow specific sanitized HTML values inside tooltip

jleifeld:
## What?

Currently tooltip can just render plain text

## Why?

We have several scenarios in the administration where we need to render some formatting markup

## How?

Add a sanitizer and render now also the markup

## Screenshots 

See visual testing screenshots


tajespasarela (approved):

packages/component-library/src/composables/useSanitizedHtml.ts
Haberkamp, line 1: Can you inline this composable to the mt-tooltip component? 

Extracting makes it public for every single component. Right, now we only use it in the tooltip component. There's no need to make it available to other components. 

It's also very unlikely that we'll use it in the near future, somewhere else.

Haberkamp (changes requested):

packages/component-library/src/composables/useSanitizedHtml.ts
jleifeld: Yeah, makes sense 👍

tajespasarela (approved):

PR #658, opened 2025-04-03
feat: add suffix option to number-field

### LINKED ISSUE #634: Add suffix option to number-field
In the near future we will need a possibility to change the suffix with a select for the number-field.
We are currently using a custom implementation of this in the Rule-Builder to change the measurement units of the rule.
This is also needed for the implementation of the imperial system in the product configuration.

Image

### Requirements
* We need to add a select at the end of the number-field which allows for switching of units (either dimension or weight units).
* List of units for dimension
  - Millimeters (mm)
  - Centimeters (cm)
  - Meters (m)
  - Inches (in)
  - Feet (ft)
  - Yards (yd)
  - They should be in the order like shown above
* List of units for weight
  - grams (g)
  - kilograms (kg)
  - ounces (oz)
  - pounds (lb)
* The values inside the input need to be converted if the unit is changed

#### To be clarified
* The incrimation of values (via arrow keys or button inside the input) can stay the same for all units to keep the complexity to a minimum?
* The min- and max-values are the same for all units to keep the complexity to a minimum?

alastair-simon:
## What?
Closes #634

## Why?
it is required in admin to be able to toggle between imperial and metric units inside the number field 

## How?
Mt-unit-field wraps number field to provide the conversion logic and select 

## Testing?
I've added some tests to ensure the value is correctly converted


Haberkamp (commented):
Bildschirmfoto 2025-04-03 um 13 25 20


Is is correct that this item doesn't have a border radius?

packages/component-library/src/components/form/_internal/mt-unit-select/mt-unit-select.vue
```diff
@@ -0,0 +1,197 @@
...
+        @click="toggleDropdown"
+        data-testid="unit-select-trigger"
+      >
+        {{ modelValue }}
+        <MtIcon name="chevron-down-xxs" size="var(--scale-size-8)" aria-hidden="true" />
+      </button>
+    </template>
+
+    <div class="mt-unit-select">
+      <ul id="unit-select-listbox" role="listbox" tabindex="-1" aria-label="Select unit">
```
Lucsy3012, line 21: Would the `Select unit` also need to be translated or how do you handle such cases in Meteor?

alastair-simon: The select unit ie. 'cm, mm...' do not in this case need to be translated as they are standardised internationally. If this component did need to be used outside of Europe of the US we would need to think about this though.

Lucsy3012: I actually meant the `aria-label="Select unit"` 😁

alastair-simon: okay thanks, good spot. I've fixed that now 😀

Lucsy3012 (dismissed):

Haberkamp (approved):

Haberkamp (dismissed):
You need to add the component to the `src/index.ts` file. Otherwise, other projects cannot use that component.

alastair-simon:
@vienthuong I believe you might have some input here regarding the units we are choosing to implement?

sydinh:
Hi @alastair-simon @vienthuong, to better serve the DB designs, I wonder if we should let the consumer freely convert the current value of the component. The only two things that the `mt-unit-field` should expose are the current value and the selected unit 🤔

sydinh:
> @vienthuong @sydinh how shall we proceed here?

Hi @alastair-simon Can the conversion function be exported from the library? We want to reuse it within the administration.

sydinh:
Hi @alastair-simon I spotted one issue while checking this component locally. The field's dropdown menu seems to have a higher index than the other parts in the admin.

index-issue

alastair-simon:
> Hi @alastair-simon I spotted one issue while checking this component locally. The field's dropdown menu seems to have a higher index than the other parts in the admin.
> 
> index-issue
Thanks @sydinh I'll take a look at that.  

> Can the conversion function be exported from the library? We want to reuse it within the administration.
I can export a composable with the conversion logic, or you could also just import the convert-units library and utilise that.

sydinh:
> > Hi @alastair-simon I spotted one issue while checking this component locally. The field's dropdown menu seems to have a higher index than the other parts in the admin.
> > index-issue
> 
> Thanks @sydinh I'll take a look at that.
> 
> > Can the conversion function be exported from the library? We want to reuse it within the administration.
> 
> I can export a composable with the conversion logic, or you could also just import the convert-units library and utilise that.

Yes, after discussing with @vienthuong, we decided to add the convert-units package in the admin.

alastair-simon:
> Hi @alastair-simon I spotted one issue while checking this component locally. The field's dropdown menu seems to have a higher index than the other parts in the admin.
> 
> index-issue

@sydinh I've added a z-index prop to fix the issue. If this is resolved, I'd appreciate if you could approve this PR.

sydinh (approved):

nguyenquocdaile (approved):

vienthuong (approved):

packages/component-library/src/components/form/mt-unit-field/mt-unit-field.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    expect(modelValueHandler).toHaveBeenCalledWith(1.094); // 1 m = 1.094 yd
+    expect(unitHandler).toHaveBeenCalledWith("yd");
+  });
+
+  it("displays different units when property measurementType is changed", async () => {
+    // ARRANGE
+    render(MtUnitField, {
+      props: {
+        modelValue: 1,
+        defaultUnit: "g" as Unit,
```
Haberkamp, line 64: Is the type cast necessary? If yes we need to change that. Otherwise, developers have to install the convert-unit package themselves too.

packages/component-library/src/components/form/mt-unit-field/mt-unit-field.vue
```diff
@@ -0,0 +1,123 @@
...
+  </div>
+</template>
+
+<script setup lang="ts">
+import { computed, watch } from "vue";
+import MtNumberField from "../mt-number-field/mt-number-field.vue";
+import MtFieldError from "../_internal/mt-field-error/mt-field-error.vue";
+import MtUnitSelect from "../_internal/mt-unit-select/mt-unit-select.vue";
+import convert from "convert-units";
+import type { Unit } from "convert-units";
```
Haberkamp, line 52: Maybe it makes more sense to just use a custom built union type with string literals so: "mm" | "cm" and so on.

That way we only allow to convert from and to units we make available.

Haberkamp (approved):

PR #678, opened 2025-04-16
feat: add CLI for creating boilerplate extension

jleifeld:


packages/create-meteor-extension/src/commands/create-meteor-extension.ts
```diff
@@ -0,0 +1,169 @@
...
+          'No template files (.ejs) found in the project structure',
+        )
+      }
+
+      // Process .ejs template files manually
+      print.info('Processing templates...')
+      for (const file of files) {
+        try {
+          const content = filesystem.read(file)
+          const processed = ejs.render(content, { name })
```
gweiermann: I'd like `{ name }` to be more prominent as a variable, like:  
```js
const templateVariables = { name };
```
If we want to add more context to templates, it will be easier to find where to do that.

packages/create-meteor-extension/__tests__/template-dependencies.test.ts
gweiermann: It's missing a vulnerability check, which you would get with `npm audit`.  
We might also just want an `npm outdated` instead of a custom implementation. There's probably a way to set package versions in `package.json.ejs` so that `npm outdated` only complains for major updates.

OliverSkroblin:
I agree with @shyim on this.

Admin development is already the part that consumes the most time – not because of writing some JS, but because getting the UX right so it actually matches what the customer expects is where most of the effort goes.

Adding additional layers just to allow “simple” JS in the admin goes in the opposite direction of what developers currently need. It increases complexity, adds more maintenance overhead, and makes the setup harder instead of easier.

jleifeld:
> But for the bootstrapping I still would except that I just have JS files inside my app/plugin and not vite whatever.
> 
> Nobody has time to manage their vite/package.json whatever especially when you have a ton of extensions inside your project and everyone has again a new setup and maybe little bit different.
> 
> Therefore I would highly recommend having something like:
> 
> ```
> npx meteor create - Create basic structure entrypoint.js some templates
> npx meteor dev - start dev server?
> npx meteor build - builds the assets
> ```
> 
> We define the required vite/whatever version inside our meteor package

I’m not sure this would work the way described.

In an app context we still need `node_modules` for dependency resolution. That means a `package.json` is required. Otherwise we can't build, start a watcher, or run a dev server, since Node resolves dependencies from `node_modules` up the directory tree.

So even if we wrap everything behind commands like:

- `npx meteor create`
- `npx meteor dev`
- `npx meteor build`

the generated project would still need a `package.json` to define and install dependencies.

The situation is different for plugins. There, the admin watcher automatically detects the plugin and starts the dev server/build process. In that case, we don't need additional setup inside the plugin itself.

So while abstracting Vite behind a CLI is nice from a DX perspective, in apps we can’t completely avoid `package.json` and `node_modules`.

Am I overlooking something here? Do you have an idea how dependency resolution and builds would work without a local `package.json` in an app setup?

shyim:
Whats the final goal of the CLI?

Just setting up a project? Then we can create also an `example-meteor-app` Git repo, which can be properly updated and you see as user also Git diffs when you need to update something. 

As user just do: `npx tiged shopware/example-meteor-app`

packages/create-meteor-extension/__tests__/template-dependencies.test.ts
gweiermann: Just to acknowledge:
If `package.json.ejs` ever gets bigger, we should rate-limit the requests.

But for now, it should be alright.

PR #686, opened 2025-04-22
fix: modal button props

fschmtt:
Adjusted the properties `variant` and `size` to match the definitions in the `mt-button` component.

### DIFF
--- a/packages/admin-sdk/src/ui/modal/index.ts
+++ b/packages/admin-sdk/src/ui/modal/index.ts
@@ -25,6 +25,6 @@ export type buttonProps = {
   method: () => void,
   label: string,
   variant?: 'primary'|'secondary'|'critical'|'action',
-  size?: 'x-small'|'small'|'large',
+  size?: 'x-small'|'small'|'default'|'large',
   square?: boolean,
 }
--- a/packages/admin-sdk/src/ui/modal/index.ts
+++ b/packages/admin-sdk/src/ui/modal/index.ts
@@ -24,7 +24,7 @@ export type uiModalClose =
 export type buttonProps = {
   method: () => void,
   label: string,
-  variant?: 'primary'|'ghost'|'danger'|'ghost-danger'|'contrast'|'context',
+  variant?: 'primary'|'secondary'|'critical'|'action',
   size?: 'x-small'|'small'|'large',
   square?: boolean,
 }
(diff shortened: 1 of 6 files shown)


SebastianFranze (commented):
@Haberkamp This can conflict with the installed version of the component library, right? But I don't see a good strategy how to couple both packages

packages/admin-sdk/src/ui/modal/index.ts
```diff
@@ -24,7 +24,7 @@ export type uiModalClose =
 export type buttonProps = {
   method: () => void,
   label: string,
-  variant?: 'primary'|'ghost'|'danger'|'ghost-danger'|'contrast'|'context',
-  size?: 'x-small'|'small'|'large',
+  variant?: 'primary'|'secondary'|'critical'|'action',
+  size?: 'x-small'|'small'|'default'|'large',
```
Haberkamp, line 28: This is a breaking change. But right now it's already broken anyway. @shopware/product-cc-framework thoughts on this?

To prevent further miss-matches, we could export the Button Props from Meteor and import them here.

Haberkamp (changes requested):
Please add a changelog using `npx changeset`

SebastianFranze (approved):

fschmtt:
> Please add a changelog using `npx changeset`

@Haberkamp I added the changelog as you asked, I'm unsure about the level, though. Should it be `major` instead of `patch`?

And, any news about exporting the button properties instead of maintaining them at multiple locations?

packages/admin-sdk/src/ui/modal/index.ts
```diff
@@ -1,4 +1,5 @@
 import { createSender } from '../../channel';
+import type { MtButtonSize, MtButtonVariant } from '@shopware-ag/meteor-component-library/src';
```
SebastianFranze, line 2: Should work, too

```suggestion
import type { MtButtonSize, MtButtonVariant } from '@shopware-ag/meteor-component-library';
```

SebastianFranze (changes requested):

SebastianFranze (approved):

jleifeld (changes requested):
This change would lead to breaking changes for older versions. To keep it backwards compatible we keep the same types and values and adjust it in the platform. 
I will do it with this PR: https://github.com/shopware/shopware/issues/9483

PR #706, opened 2025-05-09
fix: Change predicate of  isObject to "is object"

SebastianFranze:
## What?

Just a little boy scouting :v:


jleifeld:
This change seems to cause static analysis issues

PR #727, opened 2025-05-18
Update accessibility template in line with bug-report template

wannevancamp:
## What?
Update accessibility template in line with bug-report template
Change the input field to a textarea

## Why?
An input field is quite small to describe the step to reproduce an issue


Haberkamp (approved):

Haberkamp:
Thanks for creating the PR @wannevancamp. I've just merged it!

PR #730, opened 2025-05-19
feat(component-library): add button slot to mt-empty-state

Haberkamp:
## What?

This PR adds a button slot to the `mt-empty-state` component.

## Why?

Sometimes, people needs to define their own icons to buttons.

## How?

I've just added a slot, so devs can add their custom button.

I've added a wrapping div around the button, so that there's always the correct spacing between the button and the above text.


packages/component-library/src/components/layout/mt-empty-state/mt-empty-state.vue
```diff
@@ -22,16 +22,16 @@
...
-      v-if="buttonText"
-      class="mt-empty-state__button"
-      variant="primary"
-      @click="$emit('button-click')"
-    >
-      <mt-icon name="solid-plus-circle-s" />
-      {{ buttonText }}
-      <mt-icon name="solid-long-arrow-right" size="16px" />
-    </mt-button>
+    <div v-if="buttonText || $slots.button" class="mt-empty-state__button">
```
jleifeld, line 25: What is the reason for the `div`? A simple slot around the button should be enough. When the slot is filled the default button will not be rendered.

This change in DOM structure and changing the class to a different element could lead to styling issues in the administration

jleifeld (changes requested):

packages/component-library/src/components/layout/mt-empty-state/mt-empty-state.vue
jleifeld: Understandable. But problem with this is that every CSS customization of the empty state will break with this change. If you're really sure that we or our extension developer don't need to adjust things after this change, then I am fine with it

Haberkamp: This change won't break anything.

* They current way, works fine
* The new slot is optional 
* The customisation is internal not not intended to by modified by others

All existing visual tests for the empty state are the same. We've just added another one.

jleifeld (approved):

alastair-simon (approved):

PR #734, opened 2025-05-20
Fix: mt-tool-tip hover animation

alastair-simon:
## What?
Currently  meteor tool-tips have a long hover delay than others in platform, this PR fixes that. 

## Why?
The tool tips took far too long to appear, making them seem broken

## How?
- Reduced the default hide/show delay to 300ms
- Fixed broken prop inside mt-help-text
- Smoothed animation timing


packages/component-library/src/components/overlay/mt-tooltip/mt-tooltip.vue
```diff
@@ -239,11 +239,11 @@ function useSanitizedHtml(html: string): ComputedRef<string> {
 }
 
 .v-enter-active {
-  transition-timing-function: cubic-bezier(0.165, 0.84, 0.44, 1);
+  transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
 }
 
 .v-leave-active {
-  transition-timing-function: cubic-bezier(0.55, 0.085, 0.68, 0.53);
+  transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
```
Haberkamp, line 246: I would keep the old values and copy them over to platform. Instead of the other way around.

Haberkamp (changes requested):

Haberkamp (approved):

PR #746, opened 2025-05-30
feat: support async search function in the mt-select

tajespasarela:
## What?
Adds support to `mt-select` for async `searchFunction` prop

## Why?
If the search function needs to call any API, then it needs to be async.

## How?
adding a data for the search results and returning it in the computed instead of the directly the results of the search function

## Testing?
Added for a new `isPromise` util

### DIFF
--- a/packages/component-library/src/components/form/mt-select/mt-select.vue
+++ b/packages/component-library/src/components/form/mt-select/mt-select.vue
@@ -121,6 +121,7 @@ import type { PropType } from "vue";
 import { defineComponent } from "vue";
 import { debounce } from "@/utils/debounce";
 import { getPropertyValue } from "@/utils/object";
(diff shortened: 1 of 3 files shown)


packages/component-library/src/components/form/mt-select/mt-select.vue
```diff
@@ -473,6 +470,31 @@ export default defineComponent({
     valueLimit(value) {
       this.limit = value;
     },
+
+    searchTerm(newSearchTerm) {
+      this.searchResults = [];
+      if (newSearchTerm) {
+        console.log("(mt-select) Search term changed:", newSearchTerm);
```
jleifeld, line 477: Console Log left

jleifeld (approved):

PR #747, opened 2025-06-02
fix: mt-modal closing 

### LINKED ISSUE #704: [Bug]: mt-modal does not emit event when modal is closed
### Package

@meteor/component-library

### Package version

4.9.1

https://stackblitz.com/edit/vitejs-vite-yzmaksaj?file=src%2Fcomponents%2FHelloWorld.vue

### Steps to reproduce

1. Press the button to open the dialog
2. Close the dialog through the "X" icon on the top-right, by pressing ESC or by clicking the modal barrier.
3. Try to open the dialog again by pressing the button

### Actual behavior

No event is emitted when the modal is closed and the dialog cannot be re-opened.

### Expected behavior

An event (e.g. `modal-close`) should be emitted.

### System Info

```shell

```

### Any additional comments?

While porting a shopware plugin to SW 6.7 I had to replace a ` sw-dialog`  with an ` mt-dialog`  because an ` sw-dialog`  does not work with two ` datepicker`  elements inside (breaks vue when the dialog closed) but an ` mt-dialog`  does. The ` sw-dialog`  is controlled with ` v-if` and offers a `modal-close` event to reset the variable which controls the dialog. With the ` mt-dialog`  I now have to use this workaround so that it can be opened again after it was closed through the "X" icon:
``` js
onOpenDialogPressed() {
  this.showDialog = false;
  this.$nextTick(() => this.showDialog = true);
}
```

According to this migration description, mt-dialog should also emit a modal-close event.

### Code of Conduct

- [x] I have read and agree to follow Meteor's [Code of Conduct](https://github.com/shopware/meteor/blob/main/CODE_OF_CONDUCT.md)

alastair-simon:
## What?
Fixes #704. 

## Why?
The mt-modal component contained a bug where once closed it could not be reopened.

## How?
I've added an emit to mt-modal-root to ensure changes to the state isOpen are emitted

### DIFF
--- a/packages/component-library/src/components/overlay/mt-modal/mt-modal.stories.ts
+++ b/packages/component-library/src/components/overlay/mt-modal/mt-modal.stories.ts
@@ -9,7 +9,6 @@ export default {
   title: "Components/Overlay/mt-modal",
   component: MtModal,
   args: {
-    isOpen: true,
     title: "Modal Title",
     default: "Modal content",
   },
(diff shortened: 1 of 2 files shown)


Haberkamp (changes requested):
We can close this PR and the related issue. The component works as expected.

You can listen to the `change` event and update the value. 

What we can do is to provide a `v-model` compatible API.

```vue
<template>
  <MtModalRoot :is-open="isOpen" @change="isOpen = $event">
    <MtModal title="Test">
      Close this dialog with the "X" icon and try to open it again.
      <MtButton variant="primary" @click="isOpen = false">Close</MtButton>
    </MtModal>
  </MtModalRoot>
</template>
```

PR #772, opened 2025-06-14
Use watch command matching the documentation

JoshuaBehrens:
## What?

Swap watch command

## Why?

Setup tells to build upon production template, but this does not provide the composer script

## How?

The setup likely got changed from using shopware/shopware to shopware/production and here we are now

## Testing?

Follow the instructions and fail upon watching admin


jleifeld (approved):

Haberkamp (approved):
LGTM!

PR #801, opened 2025-07-10
fix(component-library): make clear button bigger in mt-select

Haberkamp:
## What?

I've made the clear icon of the `mt-select` easier to use.

## Why?

Increasing the button size improves the a11y of the select component. 

Additionally, it increases the consistency because the clear button now looks similar to the password toggle button of the password field.

## How?

I've increased the size of the icon, now it's easier to users to click on it.

I made sure that the clear button looks like the password toggle button of the `mt-password-field`

## Testing?

If the visual tests pass everything is fine.


packages/component-library/__snapshots__/interaction-tests-form-mt-select--visual-test-suffix-snap.png
Haberkamp, line 1: Looks like this was broken before this PR. I'll create a ticket for it, so we can tackle that issue.

Haberkamp: I created a ticket: https://github.com/shopware/meteor/issues/802

PR #812, opened 2025-07-15
feat: add mt-graph

Deristes:
## What?

Add a new mt-graph component that displays graphs

## Why?

We need this for multiple services

## Screenshots (optional)
image


packages/component-library/src/components/graphs/mt-graph/mt-graph-default-options.ts
```diff
@@ -0,0 +1,48 @@
...
+        stroke: {
+          color: "#2d2e32",
+          width: 1,
+          dashArray: 0,
+        },
+      },
+    },
+    grid: {
+      show: true,
+      borderColor: "#cdced4",
```
SebastianFranze, line 31: Did you try using CSS variables here? I'm not sure if we tried it :sweat_smile:

packages/component-library/src/components/graphs/mt-graph/mt-graph-default-options.ts
```diff
@@ -0,0 +1,48 @@
+import type { ApexOptions } from "apexcharts";
+
+const defaultOptions: { [key: string]: ApexOptions } = {
```
SebastianFranze, line 3: IMO we should use a factory method here instead of a singleton

```suggestion
export function getDefaultAreaOptions: ApexOptions = {
  return {
    noData: {
      [...]
    }
  };
}
```

packages/component-library/src/components/graphs/mt-graph/mt-graph.vue
```diff
@@ -0,0 +1,64 @@
...
+    :type="type"
+    :series="series"
+    :options="chartOptions"
+    :width="width"
+    :height="height"
+    :class="`mt-graph-${type}`"
+  />
+</template>
+
+<script lang="ts">
```
SebastianFranze, line 12: we usually write new components with `<script setup>`

packages/component-library/src/components/graphs/mt-graph/mt-graph.vue
```diff
@@ -0,0 +1,64 @@
...
+    },
+    height: {
+      type: [String, Number],
+    },
+  },
+
+  computed: {
+    chartOptions(): ApexOptions {
+      // @ts-ignore
+      const defaultOptionsCopy = JSON.parse(JSON.stringify(defaultOptions[this.type] || {}));
```
SebastianFranze, line 58: This is basically what we do with `deepCopyObejct` and it would be the first time somebody is using it in the project :kekw:

see https://github.com/shopware/meteor/blob/main/packages/component-library/src/utils/object.ts#L4

SebastianFranze (changes requested):

packages/component-library/package.json
```diff
@@ -73,6 +73,7 @@
     "@vuepic/vue-datepicker": "^10.0.0",
     "@vueuse/components": "^12.4.0",
     "@vueuse/core": "^13.1.0",
+    "apexcharts": "^5.2.0",
```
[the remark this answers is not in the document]
Deristes: image

No, it does not (only as dev dependency)

Also in the docs it states that you should install Apexcharts manually.

packages/component-library/src/components/graphs/mt-graph/mt-graph-default-options.ts
Deristes: Tried it, does not seem to work correctly (especially for the area gradient).

packages/component-library/src/components/graphs/mt-graph/mt-graph-default-options.ts
Deristes: Better this way?
I was struggling since i wanted to be able to support different chart types in the future.

packages/component-library/src/components/graphs/mt-graph/mt-graph.vue
Deristes: Switched it. I also created a object function to merge objects. Might be useful somewhere else as well. :smile:

packages/component-library/src/components/graphs/mt-graph/mt-graph.vue
```diff
@@ -0,0 +1,41 @@
...
+  type?: string;
+  width?: string | number;
+  height?: string | number;
+}>();
+
+const chartOptions = computed<ApexOptions>(() => {
+  const type = props.type ?? "area";
+  const options = props.options;
+  const defaultOptions = getDefaultOptions(type);
+  const defaultOptionsCopy = deepCopyObject(defaultOptions);
```
SebastianFranze, line 34: You don't need to create a copy anymore because `getDefaultOptions` always creates a new object. Btw. this was the intention behind my suggestion :)

SebastianFranze (changes requested):

SebastianFranze (dismissed):

packages/component-library/src/components/graphs/mt-graph/mt-graph.vue
```diff
@@ -0,0 +1,144 @@
...
+      chart.value.updateSeries(props.series);
+    }
+  },
+  { deep: true },
+);
+
+watch([() => props.type, () => props.width, () => props.height], init);
+</script>
+
+<style>
```
Haberkamp, line 103: If possible enabled scoped styling. That way people cannot unintentionally overwrite the styles.

Haberkamp (changes requested):
Please rename the component and every `MtGraph` reference to `MtChart` or event `MtLineChart`.

Chart is more common in English and I also think it's the correct translation from the German word "Graphen"

packages/component-library/src/components/graphs/mt-graph/mt-graph.vue
Deristes: Enabled scoped styling :)

emreimamoglu (approved):

Haberkamp (approved):

PR #852, opened 2025-08-15
feat: add slot for popover items

quando1910:
## What?
https://github.com/shopware/meteor/issues/858

## How?
Just add the new slot `:name="`popover-items__${filter.id}`"`


Haberkamp (changes requested):
Please add a changelog.

And could you show me a code example, which shows how you would use the API/new slot.

Looking at the diff, it makes sense. But it's hard to find out if the API is easy to use and makes sense.

Issue #853, opened 2025-08-18
price renderer seems useless in mt-data-table


quando1910:
### Package

@meteor/component-library

### Steps to reproduce

- Just use the setting for a column like below in mt-data-table.
```
{
      label: "checkout.subtotalLabel",
      property: "price.totalPrice",
      renderer: "price",
      rendererOptions: {
         currencyId: '........',
      },
      position: 3,
},
```

### Actual behavior

it throw the error, the whole table broken

<img width="626" height="176" alt="Image" src="https://github.com/user-attachments/assets/87c2c032-dcaf-42df-bc36-5a4f36cf65c6" />

The reason is the data-table expect the complicated price format object.

Besides, there is another bug related to the price format check in storybook, I don't see the currency symbol
<img width="1276" height="863" alt="Image" src="https://github.com/user-attachments/assets/b9d8b6e9-0858-4f11-be6f-c47139dfc21b" />

### Expected behavior

- [ ] the table should render correctly with price symbol.
- [ ] the logic of price-renderer should be more simple. it shouldn't contain too much logic to find currency, only need to receive the symbol and price value is enough.

PR #857, opened 2025-08-19
fix: make price renderer simple

quando1910:
## What?
https://github.com/shopware/meteor/issues/853

## Why?
price-renderer not working now

## How
Remove the complicate logic, price-renderer only have 1 purpose for display the price with symbol, so we only need to pass the value and ISO code, the price formula should be handled in the higher-level components.

### DIFF
--- a/.changeset/silent-icons-repeat.md
+++ b/.changeset/silent-icons-repeat.md
@@ -1,5 +1,5 @@
 ---
-"@shopware-ag/meteor-component-library": patch
+"@shopware-ag/meteor-component-library": minor
 ---
 
 add ability to show the symbol in mt-data-table cells


quando1910:
> 1. Please add a changelog
> 2. Add tests for your changes

hi @Haberkamp , I added them.

Haberkamp (dismissed):

jleifeld (changes requested):
These changes would be major breaking changes. Also without the previous structure it is not really compatible with the pricing format in Shopware anymore

quando1910:
> These changes would be major breaking changes. Also without the previous structure it is not really compatible with the pricing format in Shopware anymore

will it can be a minor changes ? because the `price-renderer` is buggy for a long time, it can not render price with currency symbol.

quando1910:
> These changes would be major breaking changes. Also without the previous structure it is not really compatible with the pricing format in Shopware anymore

I also reviewed the platform code and noticed that `mt-data-table` hasn’t been implemented yet. Instead, it relies on `sw-entity-listing`, with pricing logic handled in the parent component (e.g. sw-product-list) and then passed into the table via slot columns. Since the meteor-component-library will be integrated across multiple projects for various use cases, I believe it’s better not to embed business logic directly into a base component.

PR #875, opened 2025-08-27
fix: unit field error state

alastair-simon:
## What?
This pr fixes the bug with the error state not being visible in mt-unit-field

## Why?
Mt-unit-field should display errors in the same way as other components

## How?
I've added a missing error prop to allow the component to display the error

## Testing?
I've included an interaction test to ensure the passed error is visible


packages/component-library/src/components/form/mt-unit-field/mt-unit-field.interactive.stories.ts
```diff
@@ -50,3 +50,29 @@ export const VisualTestUnitConversion: MtUnitFieldStory = {
...
+    // Check that the error message is visible
+    const errorMessage = canvas.getByText(errorDetail);
+    expect(errorMessage).toBeVisible();
+
+    // Verify the error detail text matches exactly
+    expect(errorMessage).toHaveTextContent(errorDetail);
+
+    // Verify correct classes are applied
+    const fieldContainer = document.querySelector(".mt-field.has--error");
+    expect(fieldContainer).toBeDefined();
```
Haberkamp, line 76: We don't need to do that. Classes are an implementation detail.

The visual test already checks if the error is styled correctly.

PR #881, opened 2025-08-29
fix: Add target attribute to mt-empty-state's link

SebastianFranze:
Resolves shopware/shopware#12153

## What?

Added attribute `target` to the external link in the `mt-empty-state` component.

## Why?

* External links should be opened in a new window
* When using the empty state in an app's IFrame, links from the empty state are opened within this IFrame. This is not what should be happening


fschmtt (approved):

Haberkamp (changes requested):
We need to add a prop for that. 

There are valid cases where you don't want to open a new tab when clicking on a link. Here's a list of examples: https://css-tricks.com/use-target_blank/

.changeset/perfect-crews-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": patch
+---
+
+Added prop `linktType` to `mt-empty-state` to control the behaviour of the displayed link.
```
fschmtt, line 5:
```suggestion
Added prop `linkType` to `mt-empty-state` to control the behaviour of the displayed link.
```

SebastianFranze:
@Haberkamp, I added a prop to control the appearance and behaviour of the link. Since the default bevahiour was to open it in the same window I changed the default linkType to `internal`. This results in a visual change (the link icon is replaced with an arrow)

Should we coordinate this change with the Designteam?

interaction-tests-layout-mt-empty-state--visual-test-extended-page-snap-diff

fschmtt (approved):

PR #889, opened 2025-09-11
feat: implement badge component

alastair-simon:
## What?
This PR implements a new badge component 

## Why?
We want to unify all badges across products  

## How?
I've implemented mt-badge 

## Testing?
I've included a couple of tests but as this is a mainly a display component, more expansive tests aren't necessary.


Haberkamp (changes requested):
Please add a changeset.

And could you add an "interactive" story with visual tests with things like:

* Different colors
* Different sizes
* With and without a status indicator
* With and without icons

Haberkamp (approved):

Issue #947, opened 2025-10-29
The min/max value check for the number field can be a bit too restrictive


jc-sw-git:
### Package

@meteor/admin-sdk

### Browser

Chrome

### Operating System

Windows

### Steps to reproduce

1. Go to https://meteor-component-library.vercel.app/?path=/story/components-form-mt-number-field--default-story&args=max:80;min:8
2. Notice that the min value is set to 8 and max is set to 80
3. Enter for example the enter 70
4. Now try to edit the value to (for example) 75 by first hitting backspace and then 5

### Actual behavior

As soon as you hit backspace at step 4 of the steps to reproduce, the number will automatically be changed from 7 to 8, as 8 is the min value. Therefore you are not able to set the actually desired value. Extension partners have reported other scenarios as well. For example, min/max value are again 8-80, but in their case, when entering 7 it is automatically adjusting to 8 already. This, however, isn't something, that I could reproduce with the link provided above.

### Expected behavior

The min/max value check shouldn't be that restrictive and should either add a delay to the check or check on lost focus.

PR #960, opened 2025-11-10
fix: allow typing intermediate values below min/max during input

alastair-simon:
## What?
closes [#947](https://github.com/shopware/meteor/issues/947) 

## Why?
When mix and max values are set, the input  values clamped whilst typing. This makes for poor UX and unexpected errors.

## How?
I have set the validation to happen only when the input is blurred allowing the user to freely type within the input.

## Testing?
I have created a unit test to test this behaviour


Haberkamp (approved):
Looks good!

PR #969, opened 2025-11-18
fix(component-library): fix border-top on card footer

Haberkamp:
## What?

Removes border-top from card footer.

## Why?

Border currently invisible because of wrong css variable assignment anyway.

## How?

Removed border-top.

## Testing?

The visual tests should be enough

### DIFF
--- a/packages/component-library/.storybook/main.ts
+++ b/packages/component-library/.storybook/main.ts
@@ -1,4 +1,6 @@
 import type { StorybookConfig } from "@storybook/vue3-vite";
+import { mergeConfig } from "vite";
+import path from "path";
 
 const config: StorybookConfig = {
   stories: ["../src/**/*.mdx", "../src/**/*.stories.@(js|jsx|mjs|ts|tsx)"],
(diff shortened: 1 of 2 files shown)


packages/component-library/src/components/layout/mt-card/mt-card.vue
```diff
@@ -379,7 +379,6 @@ const cardClasses = computed(() => ({
 
   display: flex;
   padding: var(--mt-card-footer-padding);
-  border-top: 1px solid var(--color-secondary-primary-default);
```
sydinh: I don't get it. Why does removing this CSS property resolve the issue? 🤔

fabianhueske: Hi @sydinh ✌️ This depends on what the issue actually was 😄 **Currently there's no border displayed**, because the CSS variable isn't existing. That's why all things currently already using the footer in production are already working without a visible border. That's why we decided to just delete the whole line to minimize visual breaks (double borders etc.) in things already live. Just adding a border would mean that we probably need to update a lot of places in the Admin etc. And that's just not worth it.

PR #973, opened 2025-11-24
feat: implement center prop mt-empty-state

alastair-simon:
## What?
This pr re-implements the centering of mt-empty-state in a non breaking way, devs can now opt into the centering. 

## Why?
The previous implementation caused breaking changes in projects consuming meteor 

## How?
I've added the `centered` prop which is by default true but can be toggled off for centering

## Testing?
I've added a visual test for this


packages/component-library/src/components/layout/mt-empty-state/mt-empty-state.vue
```diff
@@ -51,9 +52,11 @@ withDefaults(
     linkText?: string;
     linkType?: "external" | "internal";
     buttonText?: string;
+    leftAligned?: boolean;
```
Haberkamp: Also, add this deprecated tag here as well. That way devs see it right away in their IDE.

Haberkamp (changes requested):

jleifeld (approved):

PR #979, opened 2025-11-26
Improve tree shaking behavior

jleifeld:
## What?

Fixes https://github.com/shopware/meteor/issues/978

## Testing?

See the GH action which validates and checks the bundled output files. Also check the results here: https://github.com/shopware/shopware/pull/13798

## Current result:
Bildschirmfoto 2025-12-01 um 17 01 11

Bildschirmfoto 2025-12-01 um 17 00 50

# TODO:
- improve non-direct import tree shaking (currently bundles other, unused CSS also)
- Optimize bigger components

PR #932, opened 2025-10-14
feat: add diff preview to mt-text-editor

jleifeld:
This pull request introduces a robust HTML integrity and diff review system to the `mt-text-editor` component, enhancing security and user control when editing raw HTML. The main changes include a new gate overlay that blocks WYSIWYG editing if the initial HTML would be altered, a modal for reviewing diffs when switching from code to WYSIWYG mode, and supporting utilities for HTML parsing and diffing. Documentation and storybook scenarios have also been updated to demonstrate and test these new behaviors.

https://github.com/user-attachments/assets/19bf11a2-0e05-4ccb-b019-6d6ea6e560da

https://github.com/user-attachments/assets/685489a5-5c26-453d-8991-e0e2bd8389e0

Closes https://github.com/shopware/meteor/issues/930

PR #1016, opened 2025-12-30
feat: upgrade vue-i18n to v10

bojanrajh:
## What?

Upgrades `vue-i18n` from `v9` to `v10` - both enter end of maintenance period in 1 day: https://vue-i18n.intlify.dev/guide/maintenance.html .

## Why?

Meteor component library introduced multiple breaking changes in `v4.24.0` (https://github.com/shopware/meteor/pull/979). This seems to solve one issue. There's still an issue with `mt-text-editor` emitting tailwind css reset and other CSS.

Seems related to `import "@git-diff-view/vue/styles/diff-view.css";` in https://github.com/shopware/meteor/pull/932, but we don't use `mt-text-editor` nor `mt-text-editor-diff-modal`.

Seems related to new bundling: some global CSS is auto-imported simply by calling `import { AnyFoo } from '@shopware-ag/meteor-component-library'`, without ever using the `AnyFoo`. The situation gets worse when the component is actually registered with `app.component('AnyFoo', AnyFoo)`.

## Testing?

Local `pnpm link` + `pnpm build`.

## Anything Else?

Alternatively, we can keep bundling `vue-i18n` as in this branch: https://github.com/shopware/meteor/compare/main...dx/fix-bundle

See discussion https://github.com/shopware/meteor/pull/979#discussion_r2652840732

### DIFF
--- a/packages/component-library/package.json
+++ b/packages/component-library/package.json
@@ -106,7 +106,7 @@
     "js-beautify": "^1.15.1",
     "nanoid": "^5.0.7",
     "vue-codemirror6": "^1.3.8",
-    "vue-i18n": "^9.9.1",
+    "vue-i18n": "^10.0.8",
     "vue-smooth-reflow": "^0.1.12"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


arnoldstoba:
Closed, we are looking to decouple vue-i18n from meteor in the future - #1186

PR #1058, opened 2026-02-18
fix: mt-colorpicker null handling

alastair-simon:
## What?
Handle null and undefined for MtColorpicker’s modelValue so the component no longer throws when used with no color (e.g. in admin configs or the text editor toolbar).

closes: https://github.com/shopware/shopware/issues/14683

## Why?
When modelValue was null or undefined, the component threw:
TypeError: Cannot read properties of null (reading 'string') in isColorValid

The same error when opening the picker happened because the code assumed a non-null value and did (typeof color === 'string' ? color : color.string).startsWith(...) without guarding against null/undefined. 

Custom admin configs, text editor boxed edit often pass null for “no color”, which caused crashes.

## How?
In data() and in the modelValue watcher (and when resetting in apply mode), treat null/undefined as no color.

In isColorValid, if colorValue is not a string, return false when it’s falsy or when colorValue.string is falsy before calling .startsWith().

In watcher, derive a colorString (string or color?.string) and return early if it’s falsy before any .startsWith() calls.

## Testing?
Added/updated tests for modelValue null, undefined, and "". Asserting no crash and that opening/closing the picker works.

I've also tested the fix in a custom component config in admin and it appears to have fixed the problem.


davidtraum (commented):
Looks good

PR #1102, opened 2026-04-09
fix(mt-field-label): do not disable inheritance toggle when field is …

taltholtmann:
## What?

Remove the `:disabled` binding from `mt-inheritance-switch` inside `mt-field-label`. The inheritance toggle is no longer disabled when the parent field is disabled.

## Why?

When `mt-switch` is used in a system config with sales channel inheritance (e.g. in Shopware's `sw-system-config`), the field receives `disabled: true` when the value is inherited. This `disabled` prop was forwarded through `mt-field-label` to `mt-inheritance-switch`, making the toggle button disabled — the only way to break the inheritance was blocked.

This matches the behavior of the legacy `sw-inherit-wrapper`, where the inheritance toggle was never disabled by the field's disabled state.

Related Shopware issue: https://github.com/shopware/shopware/issues/15797

## How?

Removed `:disabled="disabled"` from `mt-inheritance-switch` in `mt-field-label.vue`. The `disabled` prop remains on `mt-field-label` (no API break) but is no longer forwarded to the toggle button.

Only `mt-switch` is affected in practice — it is the only consumer that passes both `:disabled` and `:inheritance` to `mt-field-label`.

## Testing?

Added a test in `mt-switch.spec.ts`: "can be unlinked from its inherited value even when disabled" — verifies that the inheritance toggle emits `inheritance-remove` when clicked, even with `disabled: true`.

- Without fix: 1 failed, 36 passed
- With fix: 0 failed, 37 passed

## Screenshots (optional)

N/A — the toggle button visually looks the same, it just becomes clickable when the field is disabled + inherited.

## Anything Else?

`mt-field-label` is an `_internal` component. The `disabled` prop is kept for backwards compatibility but is currently unused after this change.


packages/component-library/src/components/form/_internal/mt-field-label/mt-field-label.vue
```diff
@@ -16,7 +16,6 @@
     <mt-inheritance-switch
       v-if="inheritance !== 'none'"
       :is-inherited="inheritance === 'linked'"
-      :disabled="disabled"
```
jleifeld, line 19: Wouldn't this be a break? If someone uses a `mt-field-label` and sets `disabled` to `true`, it don't work anymore

jleifeld (changes requested):

packages/component-library/src/components/form/_internal/mt-field-label/mt-field-label.vue
taltholtmann: The current behavior is a bug — when a system config field is inherited, the inheritance toggle is disabled and the user has no way to break the inheritance.

On closer inspection though, only `mt-switch` passes `:disabled` to `mt-field-label`. All other components that use `mt-field-label` don't do this: `mt-textarea`, `mt-url-field`, `mt-email-field`, `mt-datepicker`, and `mt-radio-group` — and their inheritance toggle works fine even when the field itself is disabled.

So maybe it makes more sense to fix this in `mt-switch` directly by not passing the `disabled` prop to `mt-field-label`, rather than changing `mt-field-label` itself. This could be a minor UX break for someone, but it would be consistent with how all other field components handle this. What do you think?

MartinKrzykawski (approved):

packages/component-library/src/components/form/_internal/mt-field-label/mt-field-label.vue
taltholtmann: I removed the `:disabled` `mt-label` prop now from the mt-switch, to be consistent with all the other components.

jleifeld (approved):

PR #1106, opened 2026-04-10
feat(admin-sdk): add uiSidebarActive type and sender to sidebar module

### LINKED ISSUE #1093: Enable programmatic opening sidebar for Admin SDK
### User story

I want to open a contextual sidebar from service side programmatically, so the shop users can trigger it when perform an action from a banner or st like this.

### Acceptance criteria

- [ ] The Admin SDK provides a way to open a sidebar from an extension using a locationId.
- [ ] Platform code also need to expose `uiSidebarOpen` in ExtensionAPI

quando1910:
## What?
We need to trigger sidebar open programmatically

### DIFF
--- a/docs/admin-sdk/api-reference/ui/sidebars.md
+++ b/docs/admin-sdk/api-reference/ui/sidebars.md
@@ -1,6 +1,6 @@
 # Sidebars
 
-A sidebar provides a contextual panel that displays at the right edge of the Administration window. Unlike modals, sidebars allow users to view and interact with additional content or functionality without losing context of the main interface. Sidebars should be opened in response to user interaction rather than appearing automatically. As a best practice, avoid opening sidebars without clear user context - for example, automatically displaying extension changelog sidebars immediately after login creates a poor user experience by requiring manual dismissal of each one.
+A sidebar provides a contextual panel that displays at the right edge of the Administration window. Unlike modals, sidebars allow users to view and interact with additional content or functionality without losing context of the main interface. You can open a sidebar programmatically, not only from direct UI clicks. As a best practice, still tie openings to clear user or application context: stacking many sidebars without a reason forces repeated dismissal and harms the experience.
 
 ### Add a sidebar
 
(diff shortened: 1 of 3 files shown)


jleifeld (changes requested):
Can you also update the docs? Including also since which Shopware version this feature is available

quando1910:
> Can you also update the docs? Including also since which Shopware version this feature is available

hi, I've updated it, I will assume this function can be release from 6.7.9

docs/admin-sdk/api-reference/ui/sidebars.md
```diff
@@ -60,3 +60,20 @@ sw.ui.sidebar.remove({
 | Name | Required | Description | Available at Shopware |
 | :----------- | :------- | :--------------------------------------------------------------------------------------------------------------------------------------------- | :---------------------|
 | `locationId` | true | The id of the sidebar to remove | 6.7 |
+
+### Active a sidebar
```
jleifeld: Typo: `Activate a sidebar`

docs/admin-sdk/api-reference/ui/sidebars.md
```diff
@@ -60,3 +60,20 @@ sw.ui.sidebar.remove({
...
+
+Active a sidebar that was already registered with [`add`](#add-a-sidebar). Use this when the sidebar exists but is closed or not in front. The Administration shows the panel for the given `locationId` and loads the content you associated with that id.
```
jleifeld: Same here: `Activate a. sidebar ...`

packages/admin-sdk/src/ui/sidebar/index.ts
```diff
@@ -3,6 +3,7 @@ import { createSender } from '../../channel';
 export const add = createSender('uiSidebarAdd');
 export const close = createSender('uiSidebarClose');
 export const remove = createSender('uiSidebarRemove');
+export const active = createSender('uiSidebarActive');
```
jleifeld: One question in general about the wording. It seems to be the opposite of `close`. So `open` would be more fitting. What do you thing?

jleifeld (changes requested):

packages/admin-sdk/src/ui/sidebar/index.ts
quando1910: It's ok for me, but in the platform, the function name `setActiveSidebar`, that's why I tried to use the `active` word.
Screenshot 2026-04-16 at 1 21 19 PM

jleifeld: Makes sense. Then I would call it also `uiSidebarSetActive` here. I think this is more concise. Because `uiSidebarActive` sound more like a variable or status report

jleifeld (approved):

PR #1075, opened 2026-03-16
chore: revises docs

lasomethingsomething:
## What?
Addresses https://github.com/shopware/shopware/issues/15329. 

Questions about docs architecture/location; screenshot updates; and a few substantive issues remain TBD/TBAddressed). 

## Why?
Improve installation, clarify differences between apps and plugins, provide "why" context, fix many typos and formatting issues.

PR #1109, opened 2026-04-10
docs: meteor sdk docs edits

davidtraum:
## What?

Adresses suggested docs improvements in https://github.com/shopware/meteor/pull/1075.

### DIFF
--- a/docs/admin-sdk/api-reference/data/repository.md
+++ b/docs/admin-sdk/api-reference/data/repository.md
@@ -150,14 +150,18 @@ Clones an existing entity
 ```ts
 const exampleRepository = data.repository('your_entity');
 
-const clonedEntityId = await exampleRepository.clone('theEntityIdToClone');
+const clonedEntityId = await exampleRepository.clone(
+    'theEntityIdToClone',
+    yourApiContext
+);
 ```
 
 #### Parameters
 | Name       | Required | Default | Description                                    |
 | :--------- | :------- | :------ | :--------------------------------------------- |
 | `entityId` | true     |         | The entity id which should be cloned           |
-| `context`  | false    | {}      | Change the [request context](#request-context) |
+| `context`  | true     |         | Change the [request context](#request-context) |
+| `behavior` | false    |         | Optional clone behavior passed as the third argument |
 
 #### Return value
 This method returns the id of the cloned entity.
--- a/docs/admin-sdk/api-reference/data/repository.md
+++ b/docs/admin-sdk/api-reference/data/repository.md
@@ -161,7 +161,7 @@ const clonedEntityId = await exampleRepository.clone(
 | :--------- | :------- | :------ | :--------------------------------------------- |
 | `entityId` | true     |         | The entity id which should be cloned           |
 | `context`  | true     |         | Change the [request context](#request-context) |
-| `behavior` | false    |         | Optional clone behavior passed as the third argument |
+| `behavior` | false    |         | Configure the [clone behavior](#clone-behavior) |
 
 #### Return value
 This method returns the id of the cloned entity.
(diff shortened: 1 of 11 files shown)


docs/admin-sdk/api-reference/data/repository.md
```diff
@@ -150,14 +150,18 @@ Clones an existing entity
...
+);
 ```
 
 #### Parameters
 | Name       | Required | Default | Description                                    |
 | :--------- | :------- | :------ | :--------------------------------------------- |
 | `entityId` | true     |         | The entity id which should be cloned           |
-| `context`  | false    | {}      | Change the [request context](#request-context) |
+| `context`  | true     |         | Change the [request context](#request-context) |
+| `behavior` | false    |         | Optional clone behavior passed as the third argument |
```
gweiermann: Description should include answers to the questions:
- What is a clone behavior
- What type is the parameter

docs/admin-sdk/api-reference/ui/mediaModal.md
```diff
@@ -27,7 +27,7 @@ ui.mediaModal.open({
   allowMultiSelect: false,
   fileAccept: "image/png",
   selectors: ["fileName", "id", "url"],
-  callback: ({ fileName, id, url }) => {},
+  callback: (mediaSelections) => {},
```
gweiermann: tbf, I think it's a nice way to show the type of `mediaSelections` with just destructuring it like it was before with `{ fileName, id, url }`

docs/admin-sdk/getting-started/index.md
```diff
@@ -19,7 +19,6 @@ Shopware supports two extension types: **apps** and **plugins**. Both can use th
 Apps run on an external server and communicate with Shopware through a defined API. They are the recommended approach because:
 
 - They work with **Shopware Cloud** and self-hosted instances
-- They can be distributed through the Shopware Store
 - The frontend and backend are fully decoupled from the Shopware codebase
```
gweiermann: It's missing your suggestion
> [...] to say that they are the only option that works with Shopware SaaS.

gweiermann (approved):
I've just discovered that return values all end with a semicolon, which I find quite odd. In my opinion, we should remove the semicolons.

Also, depending on the scope of this PR, not all of Jannis's comments were resolved

docs/admin-sdk/api-reference/data/repository.md
davidtraum: I have now added a separate description with an example, in the same style as `context`.

docs/admin-sdk/api-reference/ui/mediaModal.md
davidtraum: I changed it to `[{ fileName, id, url }]` to indicate that it is an array while still maintaining the destructuring.

docs/admin-sdk/api-reference/data/repository.md
gweiermann: Looks good :)

docs/admin-sdk/getting-started/index.md
davidtraum: I changed the bullet point to also mention shopware Saas

docs/admin-sdk/api-reference/data/repository.md
```diff
@@ -254,3 +254,22 @@ const exampleContext = {
...
+
+### Clone Behavior
+You can optionally change the clone behavior of the request. The clone behavior controls how the entity is duplicated on the server.
+
+Use `overwrites` to replace values in the cloned entity before it is written, for example to set a different name or other field values on the copy.
+
+Use `cloneChildren` to control whether child entities are cloned as well. This value defaults to `true`.
+
+```ts
+const exampleCloneBehavior = {
```
gweiermann, line 266: Just an idea to emphasize the type of the parameter. The actual type is probably different:
````suggestion
```ts
interface CloneBehavior {
    overwrites: object; // object that follows the same structure as the entity to overwrite its properties
    cloneChildren: boolean;
}

const example: CloneBehavior = {
````

_Edit: but we can also postpone that idea for a second iteration where we'd add a type column in the parameters table for all APIs_

davidtraum: I like this idea but would also suggest to postpone it for the next iteration and do it in the same way for all APIs as you suggested.

It would also be nice to add those types to the SDK itself in the same step, not just to the docs. Currently the clone behavior is a parameter without a type (just `any`).

jleifeld (approved):

PR #1119, opened 2026-04-13
feat: Add consent API

SebastianFranze:
## What?

Adds new namespace `consent` to the Admin-SDK

## Why?

Extensions must be able to read a consent state and request consent. Details are in the added documentation.

### DIFF
--- a/docs/admin-sdk/api-reference/consent.md
+++ b/docs/admin-sdk/api-reference/consent.md
@@ -1,0 +1,110 @@
+# Consent
+
+:::caution
+The consent API is experimental. Behavior, payloads, and availability may still change in upcoming Shopware versions.
+:::
+
+The `sw.consent` API can be used to inspect the current state of a consent and to trigger a consent request flow in the Administration.
+
+Both `sw.consent.status()` and `sw.consent.request()` return a `Promise<Consent>`.
+
+## `Consent`
+
+`Consent` is the value object returned by the consent API. It represents the current state of a consent in Shopware.
+
+### Properties
+
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `name` | `string` | Technical name of the consent. |
+| `status` | `'unset' \| 'declined' \| 'revoked' \| 'accepted'` | Current consent status. |
+| `updatedAt` | `string \| null` | Timestamp of the last consent state update. |
+| `acceptedRevision` | `string \| null` | Revision that was accepted by the user. |
+| `lastRevision` | `string \| null` | Latest known revision of the consent. |
+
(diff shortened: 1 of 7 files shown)


docs/admin-sdk/api-reference/consent.md
```diff
@@ -0,0 +1,110 @@
...
+
+```ts
+{
+  name: 'product_analytics',
+  status: 'accepted',
+  updatedAt: '2026-04-13T11:00:00.000Z',
+  acceptedRevision: '2026-03-01',
+  lastRevision: '2026-05-01',
+  isAccepted: true,
+  isStale: false,
```
bubleg: just to nitpick: acceptedRevision != lastRevision so isStale cannot be false give isAccepted is true

packages/admin-sdk/src/consent/index.ts
```diff
@@ -0,0 +1,90 @@
...
+      if (message.name !== messageOptions.consent) {
+        return Promise.resolve();
+      }
+
+      unhandle();
+      resolve(Consent.fromStatusResponse(message.consent));
+      return Promise.resolve();
+    });
+
+    createSender('consentRequest')(messageOptions).catch((reason: unknown) => {
```
bubleg: hm

so request() currently has no timeout or rejection path after the initial consentRequest message is sent 

and if the administration never emits "consentRequestResponse", the returned promise stays pending forever and the registered handler is never cleaned up

maybe add a timeout or explicit failure/cancel response so callers can recover and listeners don’t leak?

SebastianFranze: Yes i thought about this. The problem here is that the request opens a modal in the Administration with dynamic content. It could be that the user visits the privacy link and stays there for a while to read the provided information.

bubleg: fair point

my concern is maybe not the long wait itself, but that there is currently no guaranteed terminal state: if the modal is closed/navigation happens/amdministration never emits "consentRequestResponse", the promise stays pending forever

then maybe use an AbortSignal? so the caller decides when to give up, nothing fires automatically

adrianles (approved):

packages/admin-sdk/src/consent/index.ts
SebastianFranze: added an abort function: 339d29b6fa6a9c7743df84a302781781d843bd83

bubleg (approved):

bubleg:
@jleifeld can you please take a quick look

jleifeld (approved):

PR #1127, opened 2026-04-16
feat: add validate-docs-redirects script 

davidtraum:
## What?

This PR adds a script to validate the redirects for the pages of the Meteor admin SDK documentation which are listed inside `docs.yml`.
If the redirect target is invalid by accident because the file has been moved or the path is wrong the script will fail inside the static-analysis pipeline.

## Why?

Without validation it is easy to miss adding a redirect for a moved page.

## How?

The script can be called using `pnpm run validate:docs-redirects` and is also included in the static-analysis checks.

The script does two things:
- Checking the diff of the current branch to check, if files have been moved and dont have a redirect
- Checking if all current redirect targets are still valid

Other small changes in this PR:
- Adding fetch-depth: 0 is required so the full git history including origin/<base> is available in CI, allowing the validation script to diff the PR branch against the base branch to detect moved or deleted docs pages.)
- One invalid redirect in the docs was corrected

## Testing?

- Run `pnpm run validate:docs-redirects` with the current valid `docs.yml ` file. No error occurs.
- Change a redirect path in the file to a non existent target
- Run the command again -> The script exits with an error


fabianhueske:
@davidtraum Is this still needed?

davidtraum:
> David Traum (@davidtraum) Is this still needed?

Yes, it’s still relevant. I need to make a few adjustments before it’s ready. @fabianhueske

gweiermann (approved):
really nice :D

jleifeld (approved):

PR #1139, opened 2026-04-27
feat: add hint prop to form fields

### LINKED ISSUE #1062: Hint passed via prop to form components not displayed
### Package

@meteor/component-library

### Package version

4.26.0

### Browser

_No response_

### Operating System

_No response_

_No response_

### Steps to reproduce

With shopware/shopware@daf2378 the possibility was added to pass a hint via prop to the sw-field components. This was used to add hints via system config xml by passing the snippet key with the `hint` node (also added with the same commit):
```xml
<input-field type="bool">
    ...
    <hint>global.accessibility.optionWarning</hint>
</input-field>
```

With the move to meteor components, this is not working anymore.

### Actual behavior

Hint is not displayed.

### Expected behavior

Hint should be displayed.

alastair-simon:
## What?
Closes https://github.com/shopware/meteor/issues/1062

Form fields now accept an optional hint string (with an existing #hint slot still available, and the slot winning when both are used) and show that caption under the field with the standard info icon via mt-field-hint.

## Why?
Product and admin UIs need short, consistent helper text under inputs without every consumer hand-rolling markup, layout, and the info icon.

## How?
Parents expose a hint prop plus #hint, compute showFieldHint from non-empty prop text or a present hint slot (with mt-unit-field only forwarding #hint when the wrapper actually provides it), gate mt-field-hint with v-if, and keep mt-select/mt-select-base wired so the select wraps hint content the same way.

## Testing?
Existing visual tests pass


fabianhueske:
Would be nice to add a visual test with slot usage.

packages/component-library/__snapshots__/components-slider-interaction-tests--visual-test-hint-prop-snap.png
fabianhueske, line 1: This could be wrong because of the huge spacing

arnoldstoba (approved):

fabianhueske (approved):
https://github.com/shopware/meteor/issues/1292

packages/component-library/__snapshots__/components-slider-interaction-tests--visual-test-hint-prop-snap.png
alastair-simon: @fabianhueske I've reduced the spacing here

PR #1155, opened 2026-05-27
Scope diff view tailwind dep styles + add meteor specific global styles that we want to use as defaults

### LINKED ISSUE #1128: Inconsistent styling between build and watcher caused by Tailwind leakage from @git-diff-view/vue
### Package

@meteor/component-library

### Package version

_No response_

### Browser

_No response_

### Operating System

_No response_

_No response_

### Steps to reproduce

1. Use the `mt-text-editor` component in a project.
2. Ensure the diff view dependency (`@git-diff-view/vue`) is installed but not actively used.
3. Start the app in watcher mode (Vite dev server).
4. Observe styling of elements such as headings (`h1`–`h6`) and spacings.
5. Build the application for production and compare the styling.

Context: [https://shopware-ag.slack.com/archives/CE2RGNMC6/p1776343752422469](https://shopware-ag.slack.com/archives/CE2RGNMC6/p1776343752422469)

### Actual behavior

- In watcher mode, global Tailwind styles from `@git-diff-view/vue` are applied.
- These styles override our intended styles, for example heading font sizes and weights are set to inherit instead of semi-bold.
- This leads to visible differences in typography and vertical spacing.
- In production builds, these styles are not included due to Vite optimizations, resulting in different styling.

### Expected behavior

- Styling should be consistent between watcher mode and production builds.
- External dependencies used by `mt-text-editor` should not leak global styles into the host application.
- Styles should not be overridden unintentionally.

arnoldstoba:
## What?
Fixes #1128 which causes inconsistent styles due to diff-view tailwind preflight being leaked into the global styles instead of being scoped.

## Why?
Fixing unwanted behavior of global styles being added by accident.

## How?
Switching to the pure version of the css, and also still including the tailwind default variables as a fix in order to keep the exact same visuals.

## Testing?
Tested in storybook to ensure its visually the same, building the library and checking sources for any trace of global tailwind preflight styles, which are now not appearing.


fabianhueske (approved):

fabianhueske (changes requested):
Let's add line-height to the two places in the text editor.

alastair-simon (approved):

packages/component-library/src/components/assets/css/preflight.css
```diff
@@ -0,0 +1,341 @@
...
+/*
+  Set the default placeholder color to a semi-transparent version of the current text color in browsers that do not
+  crash when using `color-mix(...)` with `currentcolor`. (https://github.com/tailwindlabs/tailwindcss/issues/17194)
+*/
+@supports (not (-webkit-appearance: -apple-pay-button)) /* Not Safari */ or
+  (contain-intrinsic-size: 1px) /* Safari 17+ */ {
+  ::placeholder {
+    color: color-mix(in oklab, currentcolor 50%, transparent);
+  }
+}
```
fabianhueske: Is this relevant for us?

arnoldstoba: not necessarily, baseline new would be ios safari 26.5 and higher, this issue appears to happen in version 16.4 according to the tailwind issue report - but I cannot verify if that issue could happen in later versions. I assume we could remove this since we are relying on very modern browser versions only but it also doesn't hurt to keep it in.

packages/component-library/.storybook/preview.ts
```diff
@@ -1,6 +1,7 @@
 import type { Preview } from "@storybook/vue3";
 import "~/src/components/assets/scss/all.scss";
-import "~/src/components/assets/scss/font.scss";
+import "~/src/components/assets/css/fonts/inter.font.css";
+import { darkTheme, lightTheme } from "./shopwareTheme";
```
fabianhueske: Is this from a faulty rebase? 🤔

arnoldstoba: uhm could be, I thought you added this. I guess ill remove it then 👍🏻

fabianhueske (approved):

PR #1156, opened 2026-05-28
fix: fix mt-floating-ui transition

alastair-simon:


alastair-simon:
This pr is no longer required, so I have closed it.

PR #1270, opened 2026-07-17
feat: add visible parameter to admin-sdk tabs

socrec:
## What?
Adds an optional `visible?: boolean` field to `ui.tabs().addTabItem()`, letting an extension show or hide its own registered tab.

## Why?
Extension tabs are registered once, globally, with no way to hide them per context. So a tab always shows even on pages where it has nothing to offer.

For more context, see https://github.com/shopware/ShopwarePayments/issues/1265

## How?
- `uiTabsAddTabItem` type adds `visible?: boolean`, it rides the existing `addTabItem` message, so no channel changes.
- The Administration reads it: hides tabs with `visible: false` and upserts by `componentSectionId` (instead of duplicating it). That side ships separately in `shopware/shopware` with PR. (TBD)

## Testing?
- Consumer verified end-to-end in the Administration: a tab registered `visible: false` then toggled `true`/`false` shows and hides correctly, with no duplicate entries on re-registration.

## Screenshots (optional)
image
image

## Anything Else?
- Blocks https://github.com/shopware/shopware/pull/18403
- Blocks https://github.com/shopware/ShopwarePayments/pull/1290


gweiermann (commented):
The overall improvement is good. But I don't agree with the upsert strategy.
The verb `add` doesn't suggests upserting and it won't be consistent across other `add`-methods.

I think it would be cleaner to add a `setVisibility` method:
```js
ui.tabs('sw-order-detail').setVisibility({
    componentSectionId: 'example-order-detail-tab-content',
    visible: false
});
```

socrec:
Thank you @gweiermann 
That makes a lot more sense 👍 updated

gweiermann (approved):

daniel3010 (approved):

namdinh-92 (approved):

untilu29 (approved):

PR #1273, opened 2026-07-20
feat(component-library): make modal initial focus optional

keulinho:
## What?

Adds an `initialFocus` prop to `mt-modal`.

## Why?

Some required-decision modals should keep focus on the triggering element instead of automatically focusing the first interactive element in their content.

## How?

Passes `initialFocus: false` to the underlying focus trap when the prop is disabled. The default behavior remains unchanged.

## Testing?

- `pnpm --filter @shopware-ag/meteor-component-library exec vitest run src/components/mt-modal/mt-modal.spec.ts`
- `pnpm --filter @shopware-ag/meteor-component-library run lint:eslint` (existing warnings only)

## Screenshots (optional)

Not applicable.

## Anything Else?

Required by https://github.com/shopware/shopware/pull/18104.

### DIFF
--- a/.changeset/modal-focus-control.md
+++ b/.changeset/modal-focus-control.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": patch
+---
+
+Allow consumers to disable automatic focus when opening a modal.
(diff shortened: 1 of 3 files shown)


keulinho:
Superseded by #1275, which focuses the dialog container by default and avoids the need for a modal-specific prop.

PR #1287, opened 2026-07-23
feat: add mt-divider component

### LINKED ISSUE #1197: New component: divider
New `mt-divider` component: visual separator.

- [ ] Horizontal and vertical orientation
- [ ] Full-bleed option (extends to container edges) + default inset
- [ ] Primary and secondary color variants (`--color-border-primary` / `--color-border-secondary`)
- [ ] Stretches to container width/height
- [ ] Separator semantics (`role="separator"`/`aria-orientation`) when meaningful; decorative otherwise

arnoldstoba:
## What?

New `mt-divider` component: a horizontal or vertical separator line.

- `orientation`: `horizontal` (default) / `vertical`
- `variant`: `solid` (default) / `dashed`
- `color`: any border color token, defaults to `color-border-secondary-default`
- `alignment`: places slot content `center` (default), `left`, or `right`; the line on the aligned side is removed
- `decorative`: removes the separator semantics for purely visual lines
- `full-bleed`: stretches the divider beyond its container by the default card padding, e.g. to span the full width of a card
- optional slot for centered content (e.g. "or" in login flows, or a small action)

Closes #1197

## Why?

Sections often need a stronger visual boundary than whitespace provides; until now there was no dedicated separator component (see #1197).

## How?

- The line segments are `::before`/`::after` pseudo-elements, so slot content renders centered between two flexible segments with a fixed gap; `alignment` suppresses the segment on the aligned side.
- Accessibility follows the separator role rules: `role="separator"` with matching `aria-orientation` by default, `role="none"` when `decorative`. With slot content the role is dropped entirely, because ARIA treats a separator's children as presentational — this keeps e.g. a button inside the divider reachable for assistive technologies.
- `full-bleed` applies negative margins of the default card padding (`--scale-size-24`): inline for horizontal dividers, block for vertical ones.

## Testing?

- Unit tests for orientation, variant, color token, alignment, decorative/separator semantics, slot content accessibility, and the full-bleed class.
- Interaction test stories incl. nine visual test cases (default, dashed, brand color, with content, left/right alignment, vertical, full-bleed horizontal + vertical in a card).
- Full-bleed layout and alignment behavior verified in headless Chromium for both orientations.

## Anything Else?

- Docs page with examples included (`apps/docs`).


fabianhueske (approved):
Looks good to me. Optional thing to do is use this divider component on our start page hero.

Image

PR #1309, opened 2026-08-06
feat: Add grant permission service banner

quynhnguyen68:
## What?
Screenshot 2026-08-11 at 15 48 33


packages/component-library/src/components/_internal/mt-grant-permission-service-banner/mt-grant-permission-service-banner.vue
```diff
@@ -0,0 +1,210 @@
...
+const props = withDefaults(
+  defineProps<{
+    /**
+     * Arrangement of the banner content.
+     *
+     * "vertical" stacks icon, text and actions for narrow containers.
+     * "compact" renders the actions below the text.
+     * "wide" renders the actions next to the text.
+     */
+    layout?: "vertical" | "compact" | "wide";
```
quando1910: the banner has 3 states because it's responsive based on its width, so I don't think we allow to specific the layout here. You can remove the layout, use the [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries) to responsive based on the current component width.

quando1910 (approved):

PR #1318, opened 2026-08-11
feat: add priority option to component sections

quando1910:
## What?

Adds an optional `priority` field to `ui.componentSection.add(...)` (`uiComponentSectionRenderer`). It lets an extension control the render order when multiple extensions register a component at the same `positionId`.

- Lower values render first (`1` = topmost).
- Must be a positive number (`1` or higher); omitted or invalid values render after entries that set a `priority`.

Includes the API-reference docs update and a changeset (`minor`).

## Why?

Multiple apps/services can register UI into the same component-section `positionId` (e.g. the dashboard's `sw-dashboard__before-content`). Today they render in arrival order, which is non-deterministic because each extension's iframe boots and posts its message asynchronously — so the order changes between page refreshes, and there is no way for an extension to opt into a stable slot.

## How?

Adds `priority?: number` to the `uiComponentSectionRenderer` type. This is a purely additive, non-breaking type change — the sender already forwards the field through `createSender`, so no runtime change is needed here.

The consuming logic (validation of `priority < 1`, and the deterministic sort: services first → ascending `priority` → registration order for ties) lives in the Administration, in [shopware/shopware#19182](https://github.com/shopware/shopware/pull/19182), which depends on this field.

## Testing?

Type-only additive change; `tsc --noEmit` (`lint:types`) passes. The admin-sdk has no unit tests for `ui/*` senders (they are thin wrappers over `createSender`, and the channel is covered by `channel.spec.ts`), so no spec change is needed here. The ordering behaviour is covered by tests in the Administration PR above.

## Screenshots (optional)

n/a

## Anything Else?

Companion to [shopware/shopware#19182](https://github.com/shopware/shopware/pull/19182) — that PR consumes `priority`. Until this SDK release ships, apps cannot set `priority` in a typed way, but behaviour for existing extensions is unchanged.


docs/admin-sdk/api-reference/ui/component-sections.md
```diff
@@ -30,9 +31,12 @@ ui.componentSection.add({
 
 #### Parameters
 
-| Name        | Required | Default | Description                                    |
-| :---------- | :------- | :------ | :--------------------------------------------- |
-| `component` | true     |         | Choose the component which you want to render. |
+| Name         | Required | Default | Description                                                                                                                                                                                                                                                            |
```
jleifeld, line 34: Add also a version notice since which Shopware version this is working. Otherwise people will wonder why it doesn't have any effect

jleifeld (changes requested):

docs/admin-sdk/api-reference/ui/component-sections.md
quando1910: thanks @jleifeld , I've updated it

jleifeld (approved):

PR #1321, opened 2026-08-13
feat: add configurable payment modal options

magdakok:
## What?

Extend the Meteor payment-flow API so services can configure the shared Shopware Admin payment modal through a typed options object.

- optional `variant`, including the new `x-large` variant (PR will be added to the Shopware repo shortly to utilise this variant)
- optional `zIndex` for controlling the modal stacking order
- optional `showHeader`, `showFooter`, and `closable` overrides
- existing defaults remain unchanged: small modal, hidden header/footer, and non-closable modal
- Shopware Admin support for the `x-large` variant and custom z-index values

## Why?

The Purchase Interface needs a wider modal and must appear above the Copilot sidebar. These values were not configurable through the existing payment-flow API. Named options make the configuration clearer and allow services to control only the values they need without relying on positional arguments.

## How?

- `startPaymentFlow` accepts a typed options object while keeping the payment location and iframe lifecycle internal to Meteor.
- Meteor forwards the supported modal options through the existing `uiModalOpen` message.
- Shopware Admin stores and forwards the new modal values to `sw-modal`.
- Merchant Assistant requests the `x-large` variant with `zIndex: 2000`.

## Testing?

- Focused payment SDK tests cover the default configuration, custom variant and z-index, and visibility-option overrides.
- TypeScript validation passed.

## Anything Else?

- Older Admin versions continue to support the existing defaults, but the new width and z-index behaviour requires the matching Admin changes to be released.
- The payment modal location remains controlled by Meteor; services cannot replace it through these options.

Before (current purchase interface):
Screenshot 2026-08-13 at 18 02 55

After (+Shopware admin change https://github.com/shopware/shopware/pull/19321):
Screenshot 2026-08-13 at 18 02 47

The new width is not shown in the screenshot because it will be introduced in a future release.

### DIFF
--- a/packages/admin-sdk/src/ui/modal/index.ts
+++ b/packages/admin-sdk/src/ui/modal/index.ts
@@ -18,11 +18,12 @@ export type uiModalOpen =
     * Only simple text is supported.
     */
    textContent?: string,
-   variant?: 'default'|'small'|'large'|'full',
+   variant?: 'default'|'small'|'large'|'x-large'|'full',
    showHeader?: boolean,
    showFooter?: boolean,
    closable?: boolean,
    buttons?: buttonProps[],
+   zIndex?: number,
  }
 
 export type uiModalClose =
--- a/.changeset/stale-ants-arrive.md
+++ b/.changeset/stale-ants-arrive.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-admin-sdk": minor
+---
+
+Add additional configuration options to the Admin SDK modal API (size variant and z-index).
--- a/packages/admin-sdk/src/_private/payment/index.spec.ts
+++ b/packages/admin-sdk/src/_private/payment/index.spec.ts
@@ -389,9 +389,43 @@ describe('Private Service Payment', () => {
         showHeader: false,
         showFooter: false,
         closable: false,
+        zIndex: undefined,
       });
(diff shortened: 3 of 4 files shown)


jleifeld (approved):

davidtraum (approved):

packages/admin-sdk/src/ui/modal/index.ts
```diff
@@ -18,11 +18,12 @@ export type uiModalOpen =
     * Only simple text is supported.
     */
    textContent?: string,
-   variant?: 'default'|'small'|'large'|'full',
+   variant?: 'default'|'small'|'large'|'x-large'|'full',
    showHeader?: boolean,
    showFooter?: boolean,
    closable?: boolean,
    buttons?: buttonProps[],
+   zIndex?: number,
```
jleifeld, line 26: Can you add this to the documentation? And create a changelog entry for it? You can read how it works in the Contributing guidelines

jleifeld (changes requested):

packages/admin-sdk/src/ui/modal/index.ts
magdakok: Thank you for the info! I did this and also rebased with the freshest main :)

jleifeld (approved):

PR #1323, opened 2026-08-14
feat: remove base field

alastair-simon:
## What

Removes the internal `mt-base-field` wrapper.

- All seven consumers (`mt-text-field`, `mt-number-field`, `mt-password-field`, `mt-colorpicker`, `mt-slider`, `mt-checkbox`, `mt-select-base`) now compose the same internal primitives the already-migrated fields use — `mt-field-label`, `mt-field-error`, `mt-field-hint`, plus a new `mt-field-addition` — over their own CSS grid.
- Deleted: `mt-base-field` and `validation.mixin.ts` (only `mt-base-field` used it).

No public props, slots or events change, and no behaviour changes: this is a structural removal and replacement only.

## Why

`mt-base-field` shipped its styles **unscoped**, including `.mt-field input`, `.mt-field select` and `.mt-field textarea`. Every input nested anywhere inside a field was restyled, which is why `mt-slider` needed counter-overrides for the number inputs it nests. `mt-colorpicker` had gone a step further and defined bare global `.mt-field__addition` rules that reached every field in the application.

On top of that the wrapper forced a shared layout on components with different needs: a stacked flex chain rather than a grid, six always-rendered DOM nodes hidden with `:empty`, a required `hasFocus` prop threaded down from every consumer, and spacing that changed when an error appeared. `mt-checkbox` mounted the whole thing — bordered block, hint wrapper, 32px margin — purely to render a `<label>`, then undid it in CSS.

Three fields (`mt-textarea`, `mt-email-field`, `mt-url-field`) had already migrated to composed primitives. This finishes that job rather than inventing a new pattern.

## How

Each field owns a grid (`label / help-text`, `input`, `error`, `hint`) and composes the shared primitives. `mt-field-addition` is a compatibility shim: same class name and same pixels as the old boxed affix, but scoped instead of global. The styles `mt-base-field` shipped globally now live scoped inside each consumer, with the same values — including quirks like the `#fff` autofill background and mount-time `createId()` ids, which are reproduced rather than fixed so this PR stays a pure refactor.

**Backwards compatibility.** The legacy structural classes are kept as aliases on the new elements — `mt-field`, `mt-block-field__block`, `mt-field__label`, `mt-field__addition`, `has--error`, `has--focus`, `is--disabled`, `is--inherited`, `mt-field--small` — so existing `:deep()` overrides keep working. They are deprecated and go in the next major.

**Not changed, on purpose:** the library still has two affix components — `mt-field-addition` (legacy chrome) and `mt-field-affix` (modern) — because collapsing them visibly changes the shape of the boxed fields.

## Testing

Unit tests pass, `vue-tsc` clean, `eslint` and `stylelint` report 0 errors. One test selector was updated: a `mt-data-table` assertion depended on a redundant nested `is--disabled` that `mt-base-field` emitted; the disabled state is still on the checkbox root.

**Visual tests.** Rendering is restored pixel-for-pixel against the existing snapshots. The four remaining diffs are intentional: the inheritance stories of `mt-text-field`, `mt-number-field` and `mt-select` now render the inheritance switch through the shared `mt-field-label`, with the same compact spacing `mt-url-field`, `mt-textarea` and `mt-switch` already use.


packages/component-library/src/components/_internal/mt-select-base/mt-select-base.vue
```diff
@@ -1,25 +1,36 @@
...
-    v-bind="$attrs"
-    :disabled="disabled"
-    :has-focus="expanded"
-    :is-inherited="isInherited"
-    :is-inheritance-field="isInheritanceField"
-    :disable-inheritance-toggle="disableInheritanceToggle"
-    @inheritance-restore="$emit('inheritance-restore', $event)"
-    @inheritance-remove="$emit('inheritance-remove', $event)"
+  <div
+    class="mt-field"
```
jleifeld, line 3: Without the base field following props are missing and not working anymore:
- `helpText`
- `required`
- `copyable`

It would be good if you can add tests for this and then add these props

.changeset/tidy-fields-compose.md
```diff
@@ -0,0 +1,11 @@
+---
+"@shopware-ag/meteor-component-library": patch
```
jleifeld, line 2: Would declare it as `minor`, because otherwise NPM can autoupdate to this version and the global CSS deletions can lead to styling issues and for people using the library they will wonder why the output looks different, even without updating the depedency manually

packages/component-library/src/components/_internal/mt-base-field/mt-base-field.vue
```diff
@@ -1,496 +0,0 @@
...
-export default defineComponent({
-  name: "MtBaseField",
-
-  components: {
-    "mt-inheritance-switch": MtInheritanceSwitch,
-    "mt-help-text": MtHelpText,
-    "mt-field-copyable": MtFieldCopyable,
-  },
-
-  mixins: [MtFormFieldMixin, MtValidationMixin],
```
jleifeld, line 91: The component used the validation mixin. So all components based on this had access to the `validation` prop which is now dropping. So
- mt-text-field
- mt-number-field
- mt-password-field
- mt-colorpicker
- mt-slider
- mt-select

are now missing the important validation property

packages/component-library/src/mixins/validation.mixin.ts
```diff
@@ -1,65 +0,0 @@
-import { defineComponent } from "vue";
-import validationService from "../utils/validation";
```
jleifeld, line 2: I would also check if the validation service is somewhere used. If not, this can also be dropped

jleifeld (changes requested):

PR #1324, opened 2026-08-17
feat: Handle shopware version for Grant button

quynhnguyen68:


packages/component-library/src/components/_internal/mt-grant-permission-service-banner/mt-grant-permission-service-banner.vue
```diff
@@ -110,6 +117,11 @@ async function handleGrantPermission() {
   isLoading.value = true;
 
   try {
+    if (await compareIsShopwareVersion("<", "6.7.14.0")) {
```
quando1910: I think this block can move to the top, before isLoading, we will not trigger isLoading in this case.

PR #1325, opened 2026-08-18
feat: automate changeset in icon-kit workflow

### LINKED ISSUE #1313: Auto-generating a changeset in the icon script
### Package(s)

@meteor/icon-kit

### Current Situation

Doesn't generate a changeset file.

### Proposed Improvement

Generates one with additions, changes, deleteions, etc.

alastair-simon:
## What?
Closes #1313

Adds a `changeset` script to `@shopware-ag/meteor-icon-kit` that inspects what the Figma icon sync changed and writes a matching changeset file into `.changeset/`. The "Update icons" GitHub workflow now runs this step right after the sync, so the automated "Update icons" PR ships with its changeset already included.

## Why?

The automated icon PRs land without a changeset, so someone has to write one by hand before a release can be cut, and the icon names have to be copied over manually. That is tedious and easy to forget or get wrong. The changelog format for icon releases has been stable since v5.6.0 (a minor bump plus lists of added, modified and removed icons), so it can be generated reliably.

## How?

- **Diff via git.** After the sync, `git status --porcelain` scoped to the icons directory gives the exact set of added, modified and removed SVGs. Generated artifacts like `meta.json` and the stylesheets are filtered out, so a sync with no real icon changes writes nothing.
- **Render the existing format.** Paths become published icon names (`regular-cog`), and the file is written as a `minor` bump with one section per change kind, matching the changelog since v5.6.0.
- **Stable filename.** The file is named from a hash of the changes, so re-running the sync on the same icon set updates the existing changeset instead of creating duplicates.
- **Wiring and tests.** New `changeset` script in `package.json`, a "Generate changeset" step in `update-icons.yml`, and unit plus end-to-end tests against a throwaway git repository.

### DIFF
--- a/packages/icon-kit/package.json
+++ b/packages/icon-kit/package.json
@@ -30,6 +30,7 @@
     "lint:eslint": "eslint ./src",
     "lint:types": "tsc --noEmit",
     "start": "tsx ./src/index.ts",
+    "changeset": "tsx ./src/changeset/index.ts",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "test:unit": "vitest"
--- a/packages/icon-kit/src/changeset/changeset-file.ts
+++ b/packages/icon-kit/src/changeset/changeset-file.ts
@@ -1,0 +1,44 @@
+import {
+  ICON_CHANGE_KINDS,
+  type IconChangeKind,
+  type IconChanges,
+} from "./icon-changes.js";
+
+export type Bump = "major" | "minor" | "patch";
+
+const HEADINGS: Record<IconChangeKind, { singular: string; plural: string }> = {
+  added: { singular: "Added icon:", plural: "Added multiple icons:" },
+  modified: { singular: "Modified icon:", plural: "Modified icons:" },
+  removed: { singular: "Removed icon:", plural: "Removed icons:" },
+};
+
+/**
+ * A changeset for the icon kit, rendered in the format the icon changelog has
+ * used since v5.6.0: a bump for a single package, followed by one section per
+ * kind of change listing the affected icon names as inline code.
+ */
+export class ChangesetFile {
+  constructor(
+    private readonly packageName: string,
+    private readonly bump: Bump,
+    private readonly changes: IconChanges
+  ) {}
+
+  toString(): string {
+    const frontmatter = `---\n"${this.packageName}": ${this.bump}\n---`;
+
+    const sections = ICON_CHANGE_KINDS.flatMap((kind) => {
+      const icons = this.changes[kind];
+      if (icons.length === 0) {
+        return [];
+      }
+
+      const heading =
+        icons.length === 1 ? HEADINGS[kind].singular : HEADINGS[kind].plural;
+
+      return [[heading, ...icons.map((icon) => `\`${icon}\``)].join("\n")];
+    });
+
+    return `${[frontmatter, ...sections].join("\n\n")}\n`;
+  }
+}
--- a/packages/icon-kit/src/changeset/generate-changeset.ts
+++ b/packages/icon-kit/src/changeset/generate-changeset.ts
@@ -1,0 +1,84 @@
+import { execFileSync } from "node:child_process";
+import fs from "node:fs";
+import path from "node:path";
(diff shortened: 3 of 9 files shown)


fabianhueske (approved):

packages/icon-kit/package.json
```diff
@@ -30,6 +30,7 @@
     "lint:eslint": "eslint ./src",
     "lint:types": "tsc --noEmit",
     "start": "tsx ./src/index.ts",
+    "changeset": "tsx ./src/changeset/index.ts",
```
arnoldstoba, line 33: pnpm resolves `package.json` scripts before workspace binaries. Inside `packages/icon-kit`, `pnpm changeset` now runs `tsx ./src/changeset/index.ts` instead of `@changesets/cli`; `pnpm changeset --help` there prints "No icon changes detected" instead of the changesets help. Please rename the script to `changeset:generate` and update the step in `.github/workflows/update-icons.yml`.

packages/icon-kit/src/changeset/generate-changeset.ts
```diff
@@ -0,0 +1,88 @@
...
+  ).trim();
+
+  // git reports paths relative to the repository root, always with forward
+  // slashes, so build the prefix we match against the same way.
+  const iconsPrefix = `${path
+    .relative(repositoryRoot, iconDirectory)
+    .split(path.sep)
+    .join("/")}/`;
+
+  const changes = parseIconChanges(
```
arnoldstoba, line 53: Download failures end up here as removed icons. `src/index.ts` calls `removeDirectory(iconDirectory)` (line 28) and re-downloads every icon; per-icon failures go through `.handleError((error) => console.log(error))` (line 67) and the process exits 0. An icon whose download failed is missing from the worktree, `git status` reports it as ` D`, and this script writes it under "Removed icon". The PR is created automatically, so the changeset and the missing file reach `main` without anyone checking.

Fix belongs in the sync, not here: fail the process when an icon in `iconMap` was not written, either by rethrowing in `handleError` or by comparing the written count against `iconMap.size` before phase 2. The "Update icons" step then fails and the workflow never reaches the changeset and PR steps.

packages/icon-kit/src/changeset/generate-changeset.ts
```diff
@@ -0,0 +1,88 @@
...
+    return { changes, repositoryRoot, changesetPath: undefined };
+  }
+
+  // Derive the filename from the changes themselves so re-running the sync on
+  // an unchanged icon set updates the existing changeset instead of piling up
+  // near-duplicates.
+  const changesetPath = path.join(
+    repositoryRoot,
+    ".changeset",
+    `update-icons-${md5(changes).slice(0, 8)}.md`
```
arnoldstoba, line 79: The hash-based name is described as letting re-runs update the existing file. In the workflow every run is a fresh checkout on a new timestamped branch, so no existing file is ever present. Locally, a re-run with different changes writes a second file and leaves the first one in place (`generate-changeset.test.ts` line 167 asserts three files). A fixed name, `.changeset/update-icons.md`, gives the same result in CI and removes the `node:crypto` import, `md5()` and two tests.

packages/icon-kit/src/changeset/changeset-file.ts
```diff
@@ -0,0 +1,44 @@
...
+  modified: { singular: "Modified icon:", plural: "Modified icons:" },
+  removed: { singular: "Removed icon:", plural: "Removed icons:" },
+};
+
+/**
+ * A changeset for the icon kit, rendered in the format the icon changelog has
+ * used since v5.6.0: a bump for a single package, followed by one section per
+ * kind of change listing the affected icon names as inline code.
+ */
+export class ChangesetFile {
```
arnoldstoba, line 20: Nit: every call site is `new ChangesetFile(packageName, bump, changes).toString()`. A function `renderChangeset(packageName, bump, changes): string` does the same without the class.

arnoldstoba (changes requested):
Ran the tests, type check and lint locally (all pass) and executed the script against a deleted and an added SVG. The output matches the changelog format used since 5.6.0. Two things to change (script name, download-failure handling in the sync) and two suggestions, see inline.

PR #1328, opened 2026-08-19
fix: keep the href router-link resolves in mt-link

### LINKED ISSUE #1147: mt-link href/hover preview renders as /[object Object] on dynamic/params url structured
### Package

@meteor/component-library

### Package version

4.24.0

### Browser

Chrome

### Operating System

Windows

_No response_

### Steps to reproduce

Use a mt-link with a dynamic build route from an router path object.

Inside Shopware it self the dokuments listing is on example. Here the Example code.

```HTML
<mt-link :to="{ name: 'sw.settings.document.detail', params: { id: item.id } }">
    {% block sw_settings_document_list_columns_name_link_inner %}
        {{ item.name }}
    {% endblock %}
</mt-link>
```

### Actual behavior

The href/preview on the hover shows 
```
<domain>/[object Object]
```
Image

instead of the target URL.

### Expected behavior

The href/preview should show the accuals target url.

<doamin>admin#../../..

Effect whould be that the preview and Browser features like open in a new tab/strg + klick whould work as excpected.

iwv-stefan-m:
## What?

Fixes `mt-link` so links pointing at a route object get a real address again. At the moment they come out as `href="[object Object]"`.

Closes #1147.

## Why?

Hovering one of these links shows `/[object Object]` in the status bar, and ctrl/middle-clicking or "Open link in new tab" lands you on a broken page. Copying the link address doesn't work either.

It's a sneaky one, because a plain left-click still works — router-link handles the click itself and never looks at the href. So everything seems fine until you try to open something in a background tab, which is a pretty normal way to work through a list.

In the Shopware administration this affects 15 links across nine areas: customers, the login screen, flow builder, newsletter recipients, product streams and a few settings pages. Hovering "Forgot your password?" on the admin login screen is the quickest way to see it.

## How?

`mt-link` was setting the `href` itself. With a route object that turns into the string `[object Object]`, and since the attribute gets passed down to the `router-link` underneath, it also replaced the correct address `router-link` had already worked out.

Now `mt-link` leaves the `href` alone when it renders a `router-link`, and only sets one itself for a plain link with a string. The `to` prop accepts an object as well, which is how it was already being used.

## Testing?

Three tests added, and I made sure they actually catch the problem: the two new ones fail on `main` and pass with the fix, while the existing ten pass either way.

The full component library suite is green (706 passed, 2 skipped), and lint, types and formatting come out the same as on `main`.

## Screenshots (optional)

Nothing visual — the difference is in the link address rather than anything on screen.

## Anything Else?

Every existing `mt-link` test used `as="a"` with a plain string, so the default `router-link` case wasn't covered anywhere. That's likely how this slipped through, and it's the gap the new tests fill.

One thing I noticed but didn't touch: a `disabled` link still passes `to` down to `router-link`, which then supplies its own href, so disabled links may still be reachable by middle-click. Happy to look at that separately if you think it's worth fixing.


arnoldstoba:
Thanks for the contribution, looks good to me. I've pushed a small adjustment on top: with the `href` binding gone, a disabled router-link kept the `href` it resolved from to and became reachable again via middle-click / "Open in new tab". Disabled links now bind `href: undefined` to strip it and block clicks in a capture-phase handler (router-link's click guard skips default-prevented events), making them fully inert.

arnoldstoba (approved):

fabianhueske (approved):

PR #1335, opened 2026-08-24
fix(component-library): align unit select dropdown styling

fabianhueske:
## What?

The `mt-unit-field` unit dropdown now matches `mt-select-result-list__content`: same edge spacing (`--scale-size-4`), item padding, border radius token and neutral hover color. The disabled unit trigger no longer shows a pointer cursor.

The active item keeps its brand background and brand text, but the background now sits on the item itself so it picks up the padding and border radius.

## Screenshots

| Before | After |
| --- | --- |
|before|after|


arnoldstoba (approved):

fabianhueske:
+ Fix icon size
+ Add max-width and truncate text

alastair-simon (approved):

PR #1336, opened 2026-08-24
fix(component-library): forward mt-tabs attributes

keulinho:
## What?

Forward non-prop attributes of `MtTabs` to its real tab-list element.

## Why?

`MtTabs` renders `PriorityPlusMenu` as its root, and that component renders a slot/fragment. Vue cannot automatically inherit attributes such as `class` through that fragment, causing an extraneous-attributes warning and dropping the consumer class.

## How?

Disable automatic inheritance on `MtTabs` and bind `$attrs` to the `<div role="tablist">` that owns the `mt-tabs` class.

## Testing?

Added a regression test covering a consumer class and `data-*` attribute on the tab-list element.

## Anything Else?

This lays the Meteor foundation for [Shopware core PR #19607](https://github.com/shopware/shopware/pull/19607), which filters the Shopware-only `position-identifier` before passing remaining attributes through.


arnoldstoba:
Looks good. Just be aware @keulinho that when bumping the version in the admin this can lead to previously dead classes might now work due to the attributes being passed through. Claude tells me `sw-media-modal-v2__tabs-navigation`, `sw-media-modal-folder-settings__tabs`, `sw-flow-list__tab` and `sw-settings-search__searchable-content-tabs` might have changing styling - keep an eye out for that 👍🏻

arnoldstoba (approved):

davidtraum (approved):
LGTM — the fix is the idiomatic Vue 3 solution and matches the existing `mt-select` precedent; verified no in-repo consumer breaks.

Two non-blocking notes:
- Consumer classes passed to `MtTabs` were previously dropped entirely, so this patch can activate previously-inert consumer CSS (e.g. compound selectors like `.foo.mt-tabs`). Consider a sentence in the changeset so consumers know to audit classes they pass.
- The new test passes with `v-bind="$attrs"` alone — a `console.warn` spy would also pin the `inheritAttrs: false` half (the part that removes the reported warning).

jleifeld (approved):

PR #1340, opened 2026-08-25
fix: guard service permissions by Shopware version

quando1910:
## What?
Added Shopware version compatibility guards for Service permissions and context APIs.

## Why?
grant(), isGranted(), and isService() are only supported from Shopware 6.7.14.0. Calling them on older versions could send unsupported channel messages.

## How?
Throw clear errors when private APIs are used below 6.7.14.0.
Added legacy fallback in mt-grant-permission-service-banner using can("system_config:read").

### DIFF
--- a/packages/component-library/src/composables/useServicePermission.ts
+++ b/packages/component-library/src/composables/useServicePermission.ts
@@ -1,4 +1,13 @@
-import { context, _private, window } from "@shopware-ag/meteor-admin-sdk";
+import {
+  can,
+  compareIsShopwareVersion,
+} from "@shopware-ag/meteor-admin-sdk/es/context";
+import { isService as checkIsService } from "@shopware-ag/meteor-admin-sdk/es/_private/context";
+import {
+  grant as grantPermission,
+  isGranted,
+} from "@shopware-ag/meteor-admin-sdk/es/_private/permissions";
+import { routerPush } from "@shopware-ag/meteor-admin-sdk/es/window";
 import { asyncComputed } from "@vueuse/core";
 import { computed, ref } from "vue";
 import type { ComputedRef, Ref } from "vue";
@@ -30,7 +39,7 @@ export function useServicePermission(): UseServicePermissionReturn {
   const isLegacySWVersion = asyncComputed<boolean | null>(
     async () => {
       try {
-        return await context.compareIsShopwareVersion("<", "6.7.14.0");
+        return await compareIsShopwareVersion("<", "6.7.14.0");
       } catch {
         return null;
       }
@@ -48,7 +57,7 @@ export function useServicePermission(): UseServicePermissionReturn {
       return false;
     }
 
-    return await _private.context.isService();
+    return await checkIsService();
   }, false);
 
   const permissionGranted = asyncComputed<boolean | null>(async () => {
@@ -58,10 +67,10 @@ export function useServicePermission(): UseServicePermissionReturn {
 
     try {
       if (isLegacySWVersion.value) {
-        return await context.can("system_config:read");
+        return await can("system_config:read");
       }
 
-      return await _private.permissions.isGranted();
+      return await isGranted();
     } catch {
       return null;
     }
@@ -90,7 +99,7 @@ export function useServicePermission(): UseServicePermissionReturn {
 
     if (isLegacySWVersion.value) {
       try {
-        await window.routerPush({ path: shopwareServicePagePath });
+        await routerPush({ path: shopwareServicePagePath });
       } catch (error) {
         console.error("Error granting permission:", error);
       }
@@ -103,7 +112,7 @@ export function useServicePermission(): UseServicePermissionReturn {
     isGranting.value = true;
 
     try {
-      await _private.permissions.grant();
+      await grantPermission();
     } catch (error) {
       console.error("Error granting permission:", error);
     } finally {
--- a/packages/component-library/src/composables/useServicePermission.ts
+++ b/packages/component-library/src/composables/useServicePermission.ts
@@ -83,7 +83,7 @@ export function useServicePermission(): UseServicePermissionReturn {
     }
 
     if (isLegacySWVersion.value) {
-      return permissionGranted.value;
+      return !permissionGranted.value;
     }
 
     return isService.value && !permissionGranted.value;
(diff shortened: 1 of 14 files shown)


packages/component-library/src/composables/useServicePermission.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (
+      isLegacySWVersionEvaluating.value ||
+      isLegacySWVersion.value === null ||
+      permissionGranted.value === null
+    ) {
+      return false;
+    }
+
+    if (isLegacySWVersion.value) {
+      return !permissionGranted.value;
```
jleifeld, line 89: The banner is supposed to only appear for Shopware Services. On old Shopware versions, it can now appear for anything.

The banner asks the merchant to grant permissions for a Shopware Service. So it should only show up if two things are true: (a) this extension actually is a Service, and (b) permission isn't granted yet.

On Shopware older than 6.7.14.0 there's no way to ask "am I a Service?". That feature doesn't exist yet. Before this PR that didn't matter: the question was asked anyway, nobody answered, the call failed, and the banner just stayed hidden. Accidental, but the outcome was correct. Never shown on old versions.

This PR keeps check (b) and drops check (a). On old versions it now shows the banner whenever `can("system_config:read")` is `false`, and nothing about that condition says "this is a Service." So a plain, unrelated app that renders this banner on an old Shopware will now display "grant Shopware Services permission" to the merchant, for no reason.

There's a second thing: `can()` doesn't mean what the code needs it to mean. It reports which privileges the extension itself declared in its manifest. Not whether the merchant granted consent. So even for real Services, it's a guess standing in for the real answer, and nothing in the code explains why that particular privilege was picked.

Two ideas: make the old-version branch return `false` (keep the old "never show" behavior, just on purpose this time), or, if showing it there is genuinely wanted, say so in the PR description and find something that actually indicates "this is a Service."

packages/component-library/src/index.ts
```diff
@@ -56,6 +56,10 @@ import {
   type UseThemeOptions,
   type UseThemeReturn,
 } from "./composables/useTheme";
+import {
+  useServicePermission,
```
jleifeld, line 60: Changeset is missing

packages/component-library/src/composables/useServicePermission.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return isService.value && !permissionGranted.value;
+  });
+
+  async function grant(): Promise<void> {
```
jleifeld, line 95: If I do this:
```
try {
  await grant();
  emit("grant-success");
} catch (error) {
  emit("grant-error", error);
}
```

Then it emits success even when it is evaluating or the version is unresolvable

packages/component-library/src/composables/useServicePermission.ts
```diff
@@ -0,0 +1,131 @@
...
+  }, false);
+
+  const permissionGranted = asyncComputed<boolean | null>(async () => {
+    if (isLegacySWVersionEvaluating.value || isLegacySWVersion.value === null) {
+      return null;
+    }
+
+    try {
+      if (isLegacySWVersion.value) {
+        return await can("system_config:read");
```
jleifeld, line 70: I don't find any documentation or explanation "why" excatly this privilege stands for `Services consent was granted`. We should write the reason for this somewhere down

packages/component-library/src/composables/useServicePermission.ts
```diff
@@ -0,0 +1,131 @@
...
+  /** Whether the required permission is granted, or `null` when it cannot be resolved. */
+  permissionGranted: Ref<boolean | null>;
+  /** Whether permission-related UI should be displayed. */
+  isShowPermissionUI: ComputedRef<boolean>;
+  /** Grants native service permissions or opens the Services page on legacy Shopware versions. */
+  grant: () => Promise<void>;
+}
+
+/** Resolves and grants permissions required by a Shopware Service. */
+export function useServicePermission(): UseServicePermissionReturn {
```
jleifeld, line 32: Documentation for this composable is missing

jleifeld (changes requested):

packages/component-library/src/composables/useServicePermission.ts
quando1910: I added a comment on this line. For Shopware versions earlier than 6.7.14.0, the only way to determine whether a permission has been granted is to try making a request. Most services now include the `system_config:read` permission, and I’ll make sure that all services have this basic permission. Therefore, this check should be reliable.

packages/component-library/src/composables/useServicePermission.ts
quando1910: this composable only for our services, so I think a confluence page for how to integrate it is better than add it into public documentation. WDYT ?

packages/component-library/src/composables/useServicePermission.ts
quando1910: For older versions, yes, theoretically all apps can integrate the UI because we can't determine whether the current app is a service. However, in older versions, the component only navigates the user to the service settings page; it does not grant any permissions directly.

The permission-granting section should be available in all Shopware versions. If we hide it in older versions, each service would need to handle this manually, which is exactly the problem we're trying to solve with this shared component.

I don't see any reason why external apps would integrate this component, since it wouldn't be useful for them. It may also be impossible to completely prevent them from doing so. What do you think about adding a clear description to mt-grant-permission-service-banner explaining that it is intended for services only?

Regarding system_config:read, most services already have this permission, and I'll make sure that all services have it. Services usually need it to check the license plan, so it has effectively become a general permission.

quando1910: Screenshot 2026-08-27 at 3 19 27 PM

packages/component-library/src/index.ts
quando1910: I've updated

packages/component-library/src/composables/useServicePermission.ts
quando1910: this is the composable, so we can't use emit in here. I also do same logic as your suggestion in [component](https://github.com/shopware/meteor/pull/1340/changes#diff-cfa1c82a2168960aeb5ed6e53c792ccfdb3f73bdcc7c7ffbc5ddab3c6925a397R135)

packages/component-library/src/composables/useServicePermission.ts
jleifeld: Confluence is the wrong place because I don't think that a internal developer will look in Confluence for code documentation for Meteor stuff. I would recommend to write a detailed comment or an internal markdown file in this repository

jleifeld (approved):

PR #1344, opened 2026-08-26
fix: prevent number and unit field width collapse in intrinsic layouts

### LINKED ISSUE #1341: mt-number-field / mt-unit-field: inline-size container on the root breaks intrinsic sizing
### Package

@meteor/component-library

### Package version

since 5.3.2 (still present in 5.7.0)

### Browser

All (spec-level CSS behavior)

### Steps to reproduce

Render an `mt-number-field` inside any content-sized wrapper (flex item with auto basis, grid auto track):

```html
<div style="display: flex; width: 400px;">
  <div><mt-number-field :model-value="10" /></div> <!-- any shrink-wrapped wrapper -->
</div>
```

### Actual behavior

Since 5.3.2 the component root is an inline-size query container (for the stepper auto-hide), which zeroes its intrinsic width contribution — the wrapper collapses and the field's `width: 100%` resolves against ~0, rendering it 0px wide. `mt-unit-field` has the same container but no `width: 100%` fallback, so it collapses even without a wrapper. Downstream this collapsed the shipping price matrix in the Shopware Admin (shopware/shopware#19224) and squeezes the price fields in data grids (Advanced pricing, variant surcharges — as @fabianhueske noted in https://github.com/shopware/shopware/pull/19224#issuecomment-5409408350); the Admin carries a temporary global `container-type: normal` override in the meantime (shopware/shopware#19726).

### Expected behavior

The field keeps a usable intrinsic width inside content-sized ancestors, as before 5.3.2.

arnoldstoba:
## What?

Removes the inline-size containers from `mt-number-field` and `mt-unit-field` so the fields stop collapsing to zero width, and adds a small `min-width` floor to the inner input.

Closes #1341.

## Why?

`container-type: inline-size` (added in 5.3.0 for the stepper auto-hide) makes the fields report an intrinsic width of 0, so every shrink-to-fit context — flex rows, auto table cells, content-sized grid tracks like `sw-price-field` — collapses them (shopware/shopware#19224). Containment also pins the minimum width to any declared intrinsic value, so a `contain-intrinsic-inline-size` claim (earlier revision of this PR) inflated auto-sized tables instead: the containment cannot coexist with content-driven sizing.

## How?

- Remove `container-type`/`container-name` and both `@container` queries from `mt-number-field` and `mt-unit-field` — intrinsic sizing works natively again in every context
- `min-width: calc(var(--scale-size-32) + 2ch)` on the inner input keeps ~2 characters visible next to the stepper; on the input itself so prefix/suffix slots are unaffected
- `mt-slider` hides the steppers of its compact fields via the existing `showControls` prop (its CSS-based hiding had been dead since a 2024 class rename — the container query only papered over that)

## Testing?

Full suite green. Verified manually in the admin: advanced pricing fills its columns again, the variants surcharge matrix and shipping price matrix compress to their old widths (no table inflation), stepper behavior unchanged where grids hide them, slider shows no steppers.

## Anything Else?

The follow-up shopware MR shrinks to: the meteor version bump and the revert of the temporary global `container-type: normal` override (shopware/shopware#19726). No consumer-side `min-width` rules needed — squeezing works natively again.


daniel3010:
hi @arnoldstoba ,

I've tested this in some places like "advanced pricing" or "variant overview/modal", ... and they looks good and the stepper auto-hide still triggers correctly.

One side effect: in auto-sized data-grid columns the 200px claim widens the table. The shipping price matrix goes from 2339px to 4843px (quantity column 148px → 241px, each currency column 200px → 449px), so it needs roughly twice the horizontal scrolling.

So my question is: could the declared intrinsic width be decoupled from the query threshold?
Any single value has to be both wide enough to keep the steppers and narrow enough not to
inflate columns, and those pull against each other.

arnoldstoba:
@daniel3010 We've had a longer look at this yesterday once again. We might just go with a different solution and removing this container query altogether.

We'd keep the behavior of the `showControls` to be true by default, but since most of the input fields are `size="small"` anyways they will not disturb the admin layouts you've mentioned. I've also added a minimum width to the inner input container to ensure that there is enough space for the numbers (currently 2ch), which has been a problem previously.

With that solution we just avoid this width issue and simplify the behavior again.

fabianhueske (approved):

daniel3010:
hi @arnoldstoba, re-tested this revision everywhere we knew was broken — advanced pricing, variants overview + modal, tax inline edit — all fixed, with table and column widths same as before, so the inflation is gone too. 👍

PR #1351, opened 2026-08-31
fix(component-library): shrink mt-floating-ui inside the reference's scroll panes

gweiermann:
## What?

`mt-floating-ui` now shrinks its floating content to the space left in the scroll pane its reference element belongs to, and flips it to the other side only when that space is unusable.

## Why?

The acceptance tests fail because of the issue: https://github.com/shopware/shopware/actions/runs/33373554496/job/99430820745

A dropdown opened low inside a dialog can extend past the dialog's bottom edge and cover its buttons.

Example: In the Shopware admin's Flow Builder "Add tag" modal the open tag list hides the whole footer, so clicking where "Add action" sits toggles a tag instead of confirming the dialog.

## How?

- Adds size() and flip() middlewares.
- It calculates the parent overflow boundaries of the referenced element
- min height is 150px

## Testing?

- added tests to `mt-floating-ui-positioning.spec.ts`
- AI verified manually in a Shopware admin build across page selects, modals, data-grid context menus, the admin menu flyout and the colorpicker.

## Screenshots (optional)

**Fixed.** Flow Builder "Send email", Template list opened low in a scrolled modal body.

*Before* — the list keeps its full 250px, so the footer is hidden and "Cancel" and "Add action" cannot be clicked.

Template list overshooting the modal body and covering the modal footer

*After* — capped to the 164px left in the modal body and scrolling there, footer visible. That is three rows; with two-line entries the shrunk list does get small.

Template list shrunk to the modal body, footer with Cancel and Add action visible

**Improved, not solved.** Flow Builder "Add tag" — 47px are left for the list below the Tags field, under the 150px floor.

*Before* — the list covers the whole footer.

Tag list extending past the modal and covering its footer

*After* — the footer is reachable again, but the list changes sides and covers the Entity field above.

Tag list flipped above the Tags field, covering the Entity field, footer visible

**Unchanged where there is room.** Settings › Products, "Default Sales Channel": 476px below the field for a 153px list. Both builds render this identically, down to the byte, so one shot stands for both.

Sales channel list opening below its field on a settings page with room to spare

> The images are hosted on the disposable `screenshots/mt-floating-ui-hide` branch. Please delete that branch once these PRs are closed.

## Anything Else?

**Stacked on #1349** (`fix/floating-ui-hide-and-boundary`) and targeted at that branch, so the diff here shows only the sizing change. It should merge after #1349.

**Known limitation:** the 150px floor can still exceed the room a short modal leaves below the field — see the "Add tag" pair above.

Context menus and other popovers whose content does not scroll are never capped and keep exactly their current behaviour.

This is the acceptance failure behind shopware/shopware#18182.

### DIFF
--- a/.changeset/hungry-panthers-hide.md
+++ b/.changeset/hungry-panthers-hide.md
@@ -2,4 +2,4 @@
 "@shopware-ag/meteor-component-library": patch
 ---
 
-mt-floating-ui: hide the floating content while its reference element is fully scrolled out of view, instead of letting it float over unrelated UI
+mt-floating-ui: hide the floating content while its reference element is fully scrolled out of view, instead of letting it float over unrelated UI; consumer-supplied `floatingUiOptions.middleware` now extends the default middleware instead of replacing it
--- a/.changeset/hungry-panthers-hide.md
+++ b/.changeset/hungry-panthers-hide.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": patch
+---
+
+mt-floating-ui: hide the floating content while its reference element is fully scrolled out of view, instead of letting it float over unrelated UI; consumer-supplied `floatingUiOptions.middleware` now extends the default middleware instead of replacing it
--- a/.changeset/hungry-panthers-hide.md
+++ b/.changeset/hungry-panthers-hide.md
@@ -2,4 +2,4 @@
 "@shopware-ag/meteor-component-library": patch
 ---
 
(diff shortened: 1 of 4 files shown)


gweiermann:
Comparison between sw-popover and mt-floating-ui behavior:
- sw-popover flipped up when there was more space above
- mt-popover mostly stays below the input field and doesn't flip the same way

Why we need it:
- the floating ui covers the buttons that the acceptance tests try to click -> fails currently
- Other solution: fix the ATS that they close the dropdown before clicking the button

**sw-popover:**
image

**mt-floating-ui:**
image

gweiermann:
we came to the conclusion that this behavior is weird and decided to close it. The acceptance tests should be fixed instead.

PR #1360, opened 2026-09-03
feat: refactor datepicker custom hint to mt-field-hint

alastair-simon:
## What?

Adds a `hint` prop and `#hint` slot to `mt-datepicker`, and renders the datetime time zone hint through the shared `mt-field-hint` component. `mt-field-hint` gains an `icon` prop and a wrapper that keeps the icon at its size and centered on the first line of the text.

## Why?

mt-datepicker was the only form field without a hint caption and it needs to support custom multi-line hints. Rendering every mt-datepicker hint through `mt-field-hint` gives it the same markup and behaviour as the other fields.

## How?

- `mt-field-hint`: move typography onto the root, add an `icon` prop (default `solid-info-circle`), and wrap the icon in a `1lh`-tall flex container so it stays centered on the first line and cannot shrink. This replaces the `margin-top: calc(...)` token math. 
- Also drops a `padding` line that referenced the non-existent `--scale-size-3` token and a redundant `flex-shrink` on the inner icon.
- `mt-datepicker`: replace the custom hint markup with `mt-field-hint`. A `hint` prop or slot takes precedence over the time zone hint; the time zone hint passes `solid-clock` and keeps its `field-hint` class and test id. 
- The old `.mt-datepicker__hint-icon` class and `p` element are gone, which the minor changeset calls out.
- Ship as a minor bump for the new prop and slot, plus a patch changeset for the `mt-field-hint` alignment change.

## Testing?

- Unit specs for the `hint` prop, `#hint` slot, and the time zone hint fallback in `mt-datepicker.spec.ts`
- New `VisualTestMultiLinePropHint` story for the datepicker
- `Custom hint` and `Multi-line hint` docs stories.
- tested in admin


packages/component-library/src/components/mt-datepicker/mt-datepicker.interactive.stories.ts
```diff
@@ -442,23 +440,16 @@ export const VisualTestHelpText: MtDatepickerStory = {
   },
 };
 
-export const VisualTestLongTimeZoneHint: MtDatepickerStory = {
-  name: "Should keep the time zone hint icon size when the hint does not fit",
+export const TestDefaultTimeZoneHint: MtDatepickerStory = {
```
fabianhueske, line 448: This replaces the only story that exercised the long/wrapping time zone hint, and the snapshot for it is deleted. The new story only asserts that the text `UTC` is present, so nothing covers wrapping or icon shrinking any more.

The other ten fields that render `mt-field-hint` each have a `VisualTestMultiLinePropHint` story that uses the shared `multiLinePropHint` text and `expectHintIconAlignedWithFirstLine` (which also asserts the icon keeps its size). Could you add the same story here? This area has regressed twice now and the shared helper would have caught both cases.

fabianhueske (commented):
Reviewed the refactor end to end: unit tests, types and lint are clean, and the `1lh` wrapper is the right fix (it stays correct even when `--font-line-height-xs` is not defined, because `line-height` falls back to `normal` and `1lh` follows it). API shape matches `mt-textarea` and `mt-url-field`, and the clock icon becoming `aria-hidden` via `mt-field-hint` is a small accessibility win.

One thing I would like addressed before merge (test coverage for the datepicker hint, see inline), plus a few small ones. Note this branch conflicts with `revert/hint-icon-shrink` in `mt-datepicker.vue` and `mt-datepicker.interactive.stories.ts`, so it will need a rebase once that lands.

packages/component-library/src/components/_internal/mt-field-hint/mt-field-hint.vue
```diff
@@ -39,22 +45,30 @@ withDefaults(
   gap: var(--scale-size-4);
   padding: var(--scale-size-3);
```
fabianhueske: Pre-existing, but since this block is being reworked: `--scale-size-3` is not defined anywhere in the tokens package (the scale starts at `--scale-size-4`), and this is its only usage in the repo. The `var()` is invalid at computed-value time, so the padding resolves to `0`.

Either drop the line or switch to a real token. Leaving it means every hint in the library, now including the datepicker, shifts if the token is ever added.

packages/component-library/src/components/_internal/mt-field-hint/mt-field-hint.vue
```diff
@@ -39,22 +45,30 @@ withDefaults(
...
+ * for wrapped text the icon stays on the first line. No fixed values needed.
+ */
+.mt-field-hint__icon-wrapper {
+  flex-shrink: 0;
+  display: flex;
+  align-items: center;
+  height: 1lh;
 }
 
 .mt-field-hint__icon {
```
fabianhueske: Nit: the wrapper above is the flex item of `.mt-field-hint` and already has `flex-shrink: 0`, so this rule on the inner icon no longer does anything. Can be removed.

packages/component-library/src/components/mt-datepicker/mt-datepicker.vue
```diff
@@ -261,6 +269,16 @@ const emit = defineEmits<{
   (e: "inheritance-restore"): void;
 }>();
 
+defineSlots<{
```
fabianhueske: Nit: `defineSlots` returns the slots object, so this and the `useSlots()` call below can be a single `const slots = defineSlots<{ hint?(): void }>()` (same pattern as `mt-card.vue`), and the `useSlots` import goes away.

.changeset/datepicker-hint-prop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": minor
+---
+
+mt-datepicker: add the `hint` prop and `#hint` slot for an optional caption below the field, matching the other form fields. A custom hint replaces the default time zone hint of datetime pickers, which now uses the same shared hint styling.
```
fabianhueske, line 5: Worth one sentence on the DOM change for consumers: the time zone hint is now `span.mt-field-hint.mt-datepicker__hint > (span.mt-field-hint__icon-wrapper > icon, span.mt-field-hint__text)`. The `.mt-datepicker__hint-icon` class is gone and `.mt-datepicker__hint p` no longer matches. Nothing in this repo depends on either, but external styles might.

.changeset/center-hint-icon-without-tokens.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": patch
+---
+
+mt-field-hint, mt-datepicker: keep the hint icon vertically centered on the first line of the hint text regardless of the line-height in use. Previously the icon moved to the top of the text when the `--font-line-height-xs` token was not available.
```
fabianhueske: Small note on the stated cause, not a change request: `--font-line-height-xs` is defined at `:root` in `administration/light.css` and `dark.css`, both of which `all.scss` and the Storybook preview import. Wherever the library stylesheet is loaded the token resolves, and the old `flex-start` plus `margin-top: calc(...)` rendered identically to `align-items: center` for a single-line hint.

So the regression only shows up in a consumer that loads `primitives.css` without the administration tokens. Was that where it was observed? That would explain why no snapshot caught it, and it is also why the missing visual test above matters.

fabianhueske (changes requested):

packages/component-library/src/components/mt-datepicker/mt-datepicker.interactive.stories.ts
alastair-simon: valid change, done ✅

packages/component-library/src/components/_internal/mt-field-hint/mt-field-hint.vue
alastair-simon: valid change, done ✅

packages/component-library/src/components/_internal/mt-field-hint/mt-field-hint.vue
alastair-simon: valid change, done ✅

packages/component-library/src/components/mt-datepicker/mt-datepicker.vue
alastair-simon: valid change, done ✅

.changeset/center-hint-icon-without-tokens.md
alastair-simon: Yes that's valid point. No, it was not observed in a consumer without the administration tokens. The motivation for this PR is to allow custom multi-line hints on the datepicker, where the clock icon cannot shrink once the text is wrapped. Moving the time zone hint onto mt-field-hint fixes that. Whilst reworking the hint styles I replaced the margin-top calc with the 1lh wrapper because it keeps the icon centered on the first line without depending on the token or the icon size. The token fallback is a side benefit, not the bug I was fixing.

packages/component-library/src/components/mt-datepicker/mt-datepicker.vue
```diff
@@ -84,17 +84,18 @@
...
-      <div
-        class="mt-datepicker__hint field-hint"
-        data-testid="time-zone-hint"
-        :style="{ gridArea: 'hint' }"
-      >
-        <mt-icon name="solid-clock" class="mt-datepicker__hint-icon" size="12" />
-        <p>{{ timeZone || "UTC" }}</p>
-      </div>
-    </template>
+    <!-- @deprecated tag:v5 remove field-hint class -->
```
fabianhueske, line 87: Is this still relevant?

fabianhueske (approved):

packages/component-library/src/components/mt-datepicker/mt-datepicker.vue
alastair-simon: class 'field-hint' still exists so until removal in the major it's still relevant

PR #1368, opened 2026-09-09
fix(component-library): flip the select dropdown while the page scrolls

gweiermann:
## What?

`mt-select`, `mt-entity-select` and `mt-theme-select` now flip their dropdown while the page scrolls.

## Why?

The select result list is placed by the deprecated `v-popover` directive. That directive decides one thing in its `mounted` hook: whether the list goes below or above the field. Its scroll and resize handler only rewrites `top` and `left`.

An open dropdown therefore follows its field down the page while you scroll, and keeps the side it was given at the moment it opened. Once the space below the field is gone, the list stays below and runs off the bottom of the viewport. Closing and reopening it in the same spot places it correctly, which is the tell.

`mt-floating-ui` re-evaluates its placement on every scroll and resize. It is also what `mt-popover-deprecated` has pointed at as its replacement since it was deprecated.

## How?

- Renders the result list with `mt-floating-ui`.
- Offset pinned to `0` and z-index to `1100`, keeping the spacing and stacking the directive produced.
- Item selection travels through the result context that the active-item and keyboard listeners already use, replacing a hard-coded `this.$parent.$parent.$emit()` walk that the extra component level broke.
- Outside-click detection in `mt-select-base` also accepts the floating content.
- Drops the flip transform and the popover wrapper styles.

As a by-product the mt-entity-select has significantly improved and I also updated the screenshot snapshot in playwright for that. The floating-ui is now correctly attached to the bottom of the dropdown instead of being inside it:
image

## Testing?

1. in the administration pin a local meteor build (or wait for the bot to expose a build and pin that)
2. Build admin or start the dev watcher
3. Catalogues → Products → Add product.
4. Scroll until the Tax rate select sits near the top of the page, then expand it.
5. Scroll up, so the field slides toward the bottom of the viewport.

https://github.com/user-attachments/assets/15af73ee-e944-4b48-9047-08da037da08d

## Anything Else?

The teleported dropdown is `.mt-floating-ui__content` now, and it carries the classes passed through `popoverClasses`. Styles and end-to-end selectors matching `.mt-popover-deprecated__wrapper` or the `--placement-bottom-outside` modifier no longer apply. `mt-floating-ui` marks the resolved side with `.mt-floating-ui--top` and `.mt-floating-ui--bottom`. The changeset says the same for consumers.


packages/component-library/src/components/_internal/mt-select-base/_internal/mt-select-result-list.vue
```diff
@@ -63,13 +65,14 @@ export default defineComponent({
...
   components: {
-    "mt-popover-deprecated": MtPopoverDeprecated,
+    "mt-floating-ui": MtFloatingUi,
     "mt-icon": MtIcon,
   },
 
   provide() {
     return {
       setActiveItemIndex: this.setActiveItemIndex,
+      [MtSelectResultSelectItem]: (item: unknown) => this.$emit("item-select", item),
```
gweiermann: why do we need the provide here?

packages/component-library/src/components/_internal/mt-select-base/_internal/mt-select-result-list.vue
```diff
@@ -337,21 +340,38 @@ export default defineComponent({
...
+  right: 0;
+  bottom: 0;
+  overflow: hidden;
   pointer-events: none;
 }
 
-.mt-select-result-list,
-.mt-select-result-list-popover {
+// Child combinator, not descendant: mt-floating-ui copies the root's classes
+// onto the teleported content, where `right`/`bottom` would stretch the open list.
```
gweiermann: are these comments really needed?

packages/component-library/src/components/_internal/mt-select-base/_internal/mt-select-result-list.vue
```diff
@@ -337,21 +340,38 @@ export default defineComponent({
...
   top: 0;
   left: 0;
   right: 0;
   bottom: 0;
-  overflow: hidden;
 }
 
-.mt-select-result-list-popover .mt-popover-deprecated__wrapper {
-  width: 100%;
+// Stretched so the list anchors to the field, not to an empty inline wrapper.
```
gweiermann: i don't understand that comment

packages/component-library/src/components/_internal/mt-select-base/_internal/mt-select-result.vue
```diff
@@ -153,8 +156,7 @@ export default defineComponent({
         return;
       }
 
-      // @ts-expect-error - parent.parent should be defined
-      this.$parent.$parent.$emit("item-select", this.item);
+      this.selectItem?.(this.item);
```
gweiermann: ahh the provide just solves the weird parent traversal!

packages/component-library/src/components/_internal/mt-select-base/_internal/mt-select-result-list.vue
gweiermann, line 75: provide/inject pattern to fix the ugly `this.$parent.$parent.$emit("item-select", this.item);` down below.

Issue #1298, opened 2026-07-29
Externalize all runtime dependencies from the component-library build


arnoldstoba:
### Package(s)

@meteor/component-library

### Current Situation

The library build only externalizes three packages ([vite.config.ts](https://github.com/shopware/meteor/blob/main/packages/component-library/vite.config.ts)):

```ts
export const external = ["vue", "apexcharts", "vue-i18n"];
```

Every other runtime dependency — tiptap/prosemirror, `@git-diff-view/vue`, highlight.js, codemirror, `@vuepic/vue-datepicker`, `@vueuse/*`, `@floating-ui/*`, dompurify, etc. — is compiled **into** the published output. Measured on v5.5.0 (16.9 MB unpacked):

- `dist/esm/MtTextEditor.js` is **1.9 MB** — its only external import is `vue`; all of tiptap and prosemirror are inlined.
- `dist/esm/MtTextEditorDiffModal.js` is **1.6 MB** — `@git-diff-view/vue` and highlight.js are inlined.
- Each file is shipped twice (ESM + CJS builds), so these two components alone account for ~7 MB of the package.

This causes three problems:

1. **Consumers pay twice.** The vendor code is inside the tarball *and* installed again via `dependencies` into `node_modules`, where it sits mostly unused.
2. **Duplicate-instance risk.** A consumer that instantiates tiptap itself (e.g. to register custom extensions) gets a second prosemirror instance next to the inlined one — a known source of subtle editor bugs (plugin keys, schema identity).
3. **Hidden phantom dependencies.** The editor imports ~20 individual `@tiptap/extension-*` packages (bold, italic, document, history, …) that are **not declared** in `package.json` — they only resolve as transitive dependencies of `@tiptap/starter-kit`. This is invisible today because everything is inlined at build time.

Related inconsistencies found while auditing:

- `@tiptap/starter-kit` and `@tiptap/extension-character-count` are declared as dependencies but never imported (character counting is implemented manually in `mt-text-editor.vue`).
- `@tiptap/extension-image` is pinned to `^3.2.1` while all other tiptap packages are `^2.22.3` — a v2/v3 major mix against one `@tiptap/core`.

### Proposed Improvement

- Externalize **all** runtime dependencies: replace the allowlist with an `external` callback that treats every non-relative, non-aliased import as external.
- Declare every directly imported package in `dependencies` (in particular the ~20 individual `@tiptap/extension-*` packages), so the externalized imports resolve under strict package managers (pnpm).
[...]

PR #1371, opened 2026-09-10
fix(component-library): load mt-datepicker locales through static imports

shyim:
## What?

Fixes the `mt-datepicker` `locale` prop not applying in production builds: the datepicker always showed English month and weekday names no matter which locale was passed.

## Why?

The date-fns locale was loaded through a runtime-constructed dynamic import:

```ts
await import(/* @vite-ignore */ `date-fns/locale/${path}`)
```

Bundlers cannot statically analyze that specifier. In consumer production builds (Vite, webpack) the import rejects at runtime, the `catch` swallows it, and the component silently falls back to `enUS`. It only ever worked in dev/test environments where the import is resolved by the dev server.

A symptom visible in this repo: the `should display the correct locale when a locale string is given` spec failed consistently on `main` (expected `juil.`, got `Jul`).

## How?

- `date-fns-locale.ts` now uses an explicit loader map covering all 95 date-fns v4 locales, each with a literal `import("date-fns/locale/xx")` specifier that any bundler can statically analyze.
- `date-fns` is now external in the library build (like `vue` and `vue-i18n` already are). The dist output keeps the literal `import("date-fns/locale/xx")` specifiers, so the **consumer's** bundler code-splits each locale into a lazy chunk that is only loaded when needed — no locale data is shipped inside the package. `date-fns` remains a declared dependency, so consumers need no setup change. Side effect: the datepicker chunk shrinks (412 kB → 316 kB unminified) because date-fns core is no longer bundled either.
- Locale lookup is now case-insensitive (`DE`, `en-gb`, … all work); unknown or unsafe paths still return `null` and fall back to `enUS`.
- The existing module-shape resolver from #1161 is unchanged.
- `.size-limit.ts` ignores `date-fns/*` so the lazily loaded locales are not counted (esbuild only externalizes subpaths via wildcard).

## Testing?

- Fixed the race in `mt-datepicker.spec.ts`: the locale swap is asynchronous, so the spec now waits until the French month label is applied instead of asserting immediately after the menu opens.
- Added `date-fns-locale.spec.ts` coverage for case-insensitive imports and unknown locale paths.
- Full component-library suite passes: 55 files, 781 tests.
- `lint:types` (vue-tsc), ESLint and Prettier pass.
- `size-limit` passes: `index.js` is 915 kB against the 950 kB limit (was 1020 kB when the locales were bundled into dist).
- Verified the Vite library build emits analyzable `import("date-fns/locale/xx")` specifiers in both ESM and CJS outputs.

## Anything Else?

Includes a patch changeset for `@shopware-ag/meteor-component-library`.

<div><a href="https://cursor.com/agents/bc-f228a536-319b-47e7-91b5-e40ee2062a4d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f228a536-319b-47e7-91b5-e40ee2062a4d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/datepicker-locale-loader-map.md
+++ b/.changeset/datepicker-locale-loader-map.md
@@ -2,4 +2,4 @@
 "@shopware-ag/meteor-component-library": patch
 ---
 
-mt-datepicker: fix the `locale` prop not applying in production builds. The date-fns locale was loaded through a runtime-constructed dynamic import that bundlers cannot statically analyze, so the import failed in consumer builds and the datepicker silently fell back to English month and weekday names. Locales are now loaded through statically analyzable imports, so every date-fns locale works in Vite, webpack and Rollup builds. Locale lookup is also case-insensitive now (e.g. `DE` or `en-gb` work).
+mt-datepicker: fix the `locale` prop not applying in production builds. The date-fns locale was loaded through a runtime-constructed dynamic import that bundlers cannot statically analyze, so the import failed in consumer builds and the datepicker silently fell back to English month and weekday names. Locales are now loaded through statically analyzable imports and `date-fns` is no longer bundled into the package, so the consumer's bundler code-splits each locale into a lazy chunk that is only loaded when needed. Locale lookup is also case-insensitive now (e.g. `DE` or `en-gb` work).
--- a/.changeset/datepicker-locale-loader-map.md
+++ b/.changeset/datepicker-locale-loader-map.md
@@ -1,0 +1,5 @@
+---
+"@shopware-ag/meteor-component-library": patch
+---
+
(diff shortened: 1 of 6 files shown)


arnoldstoba:
The fix seems good and logical, but we've bundled the entire "dependency externalization" topic into an issue that we would like to tackle as a whole in the next major release https://github.com/shopware/meteor/issues/1298.

shyim:
we can also close this. my jest suite fails just always locally

PR #1376, opened 2026-09-16
fix(component-library): center tags and the load-more button in mt-select

### LINKED ISSUE #20588: mt-select-result-list contains unnecessary new line
### Shopware Version

6.7.14.0

### Affected area / extension

Platform(Default)

### Actual behaviour

With SwagAnalytics installed in platform there are some visual issues in the filter area of Swag Analytics. It seems the meteor components needs some adjustments. 

- The mt-select-result-list within the Filter contains unnecessary new line (see image)
- Number placeholder has wrong alignment (see image)

Image

### Expected behaviour

- No new line for Input filed
- numbers are aligned with text

### How to reproduce

1. Install Shopware and Swag Analytics latest version (v2.12 currently)
2. Open Analytics app
3. Open Filter view
4. Choose different Sales Channels

arnoldstoba:
## What?

Centers the selected tags, the "+N" load-more button and the search input inside `mt-select` for the default and the small size. `size="small"` now renders the same 32px small select as the `small` prop.

## Why?

The tags sat 1px too low with their text hanging below center, the gap between tags was 12px because of two stacked margins, and the "+N" button sat at the top of the row. Small selects had two code paths that rendered differently.

## How?

- The selection list owns all spacing with one `gap` and block padding, every item is a centered flex row, and the input no longer dictates the row height.
- `mt-label` centers its text with flex and lost its outer margin. The medium size used by small selects is 20px tall with `2xs` text.
- `mt-select` treats `small` and `size="small"` the same.
- Dead rules for the never-rendered `--single` class and the unreachable `--medium` field size are removed.

## Testing?

- Unit tests, type check, eslint, stylelint and prettier pass.
- Five new visual stories cover default and small load-more, small multi-select, the `size="small"` path and small wrapping. Snapshots need to be regenerated with the Visual tests workflow.

## Anything Else?

The admin's per-page select in `sw-pagination` becomes truly small with this change; its workaround rules in `sw-pagination.scss` can be removed on the platform side once this version is consumed. Also fixes https://github.com/shopware/shopware/issues/20588.


packages/component-library/src/components/_internal/mt-select-base/_internal/mt-select-selection-list.vue
```diff
@@ -312,71 +308,66 @@ export default defineComponent({
 </script>
 
 <style>
```
alastair-simon, line 322: I think we can remove these comments, fairly self-explanatory

alastair-simon (approved):

